PR #2, opened 2017-05-25
Contributed Project Fuse source code

trcox:
Signed-off-by: Tyler Cox <tyler_cox@dell.com>

### DIFF
--- a/jenkins_jobs.ini
+++ b/jenkins_jobs.ini
@@ -1,0 +1,8 @@
+[jenkins]
+user=dell
+password=e9a41252f2789e57ac03624509febd09
+url=http://localhost:8080
+query_plugins_info=False
+
+## TODO ##
+##add jenkins user to credentials
--- a/jobs/global-defaults.yaml
+++ b/jobs/global-defaults.yaml
@@ -1,0 +1,12 @@
+- defaults:
+    name: global
+    build-days-to-keep: 30
+    build-num-to-keep: 40
+    build-artifact-days-to-keep: -1
+    build-artifact-num-to-keep: 5
+    # Timeout in minutes
+    build-timeout: 360
+    repos-cred-id:  96ac14d9-645d-4f92-9dfc-2d0d57932ab8
+
+## TODO ##
+# create repos credentials in Jenkins and put ID here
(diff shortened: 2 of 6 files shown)


jobs/java-maven-template.yaml
```diff
@@ -0,0 +1,101 @@
+# Library and Micro Service build creation
+
+# Template for all Java libraries
+- job-template:
+    name: '{name}-library'
+    display-name: '{name}-library'
+    description: '{description}
+
+'
+    project-type: maven
```
tykeal, line 10: We do not allow 'project-type: maven' in our build environment. The templates must use a freestyle type.

jenkins_jobs.ini
```diff
@@ -0,0 +1,8 @@
+[jenkins]
```
tykeal, line 1: This entire file needs to go away. We provide an example jenkins.ini file and our .gitignore is configure to ignore jenkins.ini files themselves as they hold credentials.

jobs/global-defaults.yaml
```diff
@@ -0,0 +1,12 @@
+- defaults:
```
tykeal, line 1: We already provide a defaults file and his one runs afoul of several problems we've generally worked around

tykeal (changes requested):
Hmm... I started going through this and marking things up but realized that there's a lot wrong with it.

First you have files that hold credentials. We never post do this, you're going to have to do a squash commit to your PR to get rid of them as well as remove them from the history. Besides, none of those credentials are valid in our environment.

Second, all the job definitions are in jobs directory instead of the JJB directory that is already part of the project. We only pull job definitions out of that directory

Third, all of these jobs are pointing at the bitbucket repositories.

Fourth, I don't see any verify / merge job distinctions here. Pretty much all of the jobs presently don't hold with our standard ideas of how the code flow should work:

PR triggers a verification job
Merge triggers a merge job which produces artifacts

Fifth, the jobs are all defined in one file / location and not broken up a more granular fashion. We generally recommend that projects (or specific project groups) use the following structure for their jobs:

jjb/<project>/<project>.yaml
jjb/<project>/<project-additions>.yaml

Given the number of problems I'm seeing I'll see if I can work up something that is more in line with our normal flows. Going off in a direction that is significantly different from our other supported projects makes it very hard for the LF Release Engineering team to support the jobs.

JPWKU:
recheck

JPWKU:
recheck

JPWKU:
recheck

PR #3, opened 2017-05-30
Configure EdgeX specific Jenkins minion scripts

tykeal:
Our build minions need start scripts for when they care coming online.
While we were initially starting the project we leveraged the start
scripts of a different project that The Linux Foundation supports. It's
best if each of the projects are using scripts that are more customized
to their needs.


JPWKU (approved):
LGTM

PR #6, opened 2017-06-13
Update global-jjb to fix for merge jobs

tykeal:
While working on updating to the v0.3.x series of global-jjb it was
discovered that the merge jobs require fully rendered SCM URLs instead
of URLs that contain variables. As a new patch release hasn't come out
we're going to point directly at the already merged fix.


zxiiro (approved):
+1

PR #15, opened 2017-07-12
Update global-jjb to 0.4.3 and GitHub

tykeal:
The Linux Foundation has started to mirror global-jjb to GitHub. As
such, it's a good idea to switch the submodule to use this repo instead
of the source Gerrit as it de-couples JJB management from the upstream
repo allowing us to do maintenance when we need to without coordinating
with all projects.

Secondly, there have been several updates to global-jjb so it's time to
pull in the latest changes as well.


JPWKU (approved):
LGTM

PR #17, opened 2017-07-24
Cleanup packer template for docker

tykeal:
The docker template was copied over from OpenDaylight and not updated to
be a single OpenStack configuration when this was done

### DIFF
--- a/packer/templates/docker.json
+++ b/packer/templates/docker.json
@@ -1,53 +1,30 @@
 {
   "variables": {
-    "public_base_image": null,
-    "public_network": null,
-    "public_cloud_user": null,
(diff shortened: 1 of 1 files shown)


JPWKU:
Not real sure what you are trying to say with the commit message.  I think it's ```Updating to a single OpenStack configuration``` ?

tykeal:
If you look at the changes, you'll see that the template goes from building two instances in two different clouds, to just building a single instance in a single cloud. OpenDaylight is a bit of a special case because they have to build in two different clouds anytime they're building a new image.

JPWKU (approved):
LGTM

PR #22, opened 2017-07-26
Add docker verify and merge job templates

tykeal:
Adds templates for doing verify and merge of docker components.

* Job templates allow for pre-build script to be passed to the job
* Job templates allow for post-build script to be passed to the job
* Post-build script is executed before the push to the local docker
  registry

### DIFF
--- a/jjb/edgex-templates-docker.yaml
+++ b/jjb/edgex-templates-docker.yaml
@@ -1,0 +1,167 @@
+---
+###########
+# ANCHORS #
+###########
+- docker_job_boiler_plate: &docker_job_boiler_plate
+    name: docker_job-boiler-plate
+
+    project-type: freestyle
+    node: '{build-node}'
+
+    ######################
+    # Default parameters #
+    ######################
+
+    branch: master
+    submodule-recursive: true
+
+    # default params for docker jobs
+    pre_build_script: ''
+    post_build_script: ''
+    docker_root: '$WORKSPACE'
+    docker_build_args: ''
+    docker_name: 'unnamed'
+    # Default to LF standard 'snapshots' docker registry
+    docker_registry: '$DOCKER_REGISTRY:10003'
+    docker_tag: 'latest'
+
+    #####################
+    # Job Configuration #
+    #####################
+
+    properties:
+      - lf-infra-properties:
+          build-days-to-keep: '{build-days-to-keep}'
+      - github:
+          url: '{git-url}/{github-org}/{project}'
+
+    parameters:
+      - lf-infra-parameters:
+          branch: '{branch}'
+          project: '{project}'
+          stream: '{stream}'
+
+    wrappers:
+      - lf-infra-wrappers:
+          build-timeout: '{build-timeout}'
+          jenkins-ssh-credential: '{jenkins-ssh-credential}'
+
+    publishers:
+      - lf-infra-publish
+
+- docker_verify_boiler_plate: &docker_verify_boiler_plate
+    name: docker_verify_boiler_plate
+
+    concurrent: true
+
+    scm:
+      - lf-infra-github-scm:
+          url: '{git-clone-url}{github-org}/{project}'
+          refspec: '+refs/pull/*:refs/remotes/origin/pr/*'
+          branch: '{branch}'
+          submodule-recursive: '{submodule-recursive}'
+          choosing-strategy: default
+          jenkins-ssh-credential: '{jenkins-ssh-credential}'
+
+    triggers:
+      - github-pull-request:
+          trigger-phrase: '^recheck$'
+          only-trigger-phrase: false
+          status-context: '{project} Verify'
+          permit-all: true
+          github-hooks: true
+          auto-close-on-fail: false
+
+- docker_merge_boiler_plate: &docker_merge_boiler_plate
+    name: docker_merge_boiler_plate
+
+    scm:
+      - lf-infra-github-scm:
+          url: '{git-clone-url}{github-org}/{project}'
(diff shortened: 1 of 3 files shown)


tykeal:
NOTE: This is a WIP and does not currently have the verify job, even though the commit message states that it does. I just wanted to get this out here for comment on the current methodology.

Current tests can be seen at: https://jenkins.edgexfoundry.org/sandbox/job/docker-core-command-master-merge-docker/

shell/docker-build.sh
```diff
@@ -0,0 +1,26 @@
...
+
+# Build the docker image
+
+# Allow word splitting
+# shellcheck disable=SC2086
+docker build $DOCKER_ARGS . | tee "$WORKSPACE/docker_build_log.txt"
+DOCKER_IMAGE=$(grep -Po 'Successfully built \K[^ ]+' \
+    "$WORKSPACE/docker_build_log.txt")
+
+IMAGE_NAME="$DOCKERREGISTRY/$DOCKER_NAME:$DOCKER_TAG"
```
zxiiro: Is DOCKERREGISTERY missing the underscore on purpose?

zxiiro (commented):
Seems generally fine although I'm not familiar with the Docker build stuff.

shell/docker-build.sh
tykeal: Yes, and where I had it in my scripts before I did some re-arranging I actually had a comment as to why. I'll get that in again before I push an updated one with the verify template as well.

tykeal: Actually, the comment for it is below this line. I didn't originally have the variable being set as I only needed the combined value once.

zxiiro: Ah missed that. I usually look above for documentation rather than below.

jjb/edgex-templates-docker.yaml
```diff
@@ -0,0 +1,141 @@
...
+
+    triggers:
+      - lf-infra-github-pr-trigger:
+          trigger-phrase: '^recheck$'
+          only-trigger-phrase: false
+          status-context: '{project} Verify'
+          permit-all: true
+          github-hooks: true
+          org-list:
+            - '{github-org}'
```
JPWKU: I think if you do this you will pass a list item to a list item.  The actual macro has this parameter set just like this.  
So perhaps it should be `org-list: '{github-org}'` ....a simple string.

JPWKU: Also, this only lets members of that org actually do the verify instead of leaving it completely open.

shell/docker-push.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+# Ensure we fail the job if any steps fail
+set -e -o pipefail
+
+# Push our image to wherever it's defined to go by the name
+docker push "$DOCKER_IMAGE"
```
JPWKU, line 6: Perhaps a `docker rmi -f $DOCKER_IMAGE` on a successful push?

JPWKU: alternatively we could set a cron to clean up old images.

tykeal: No reason to, this is being done on a single shot instance. It will automatically get cleaned up as soon as the job is done.

jjb/edgex-templates-docker.yaml
tykeal: Hmm... I'll have to look at how the jjb verify is setup. It's going to need to be similar, and flexible

tykeal:
Latest version now has both verify and merge templates. The only difference between the two is that the merge template pushes the docker image as the last build step.

JPWKU:
I thought we did have the ghprb trigger working on remerge and push events
Or maybe it was just push events.

JPWKU:
We did have it working on push events now...I remember picking the wrong event to trigger on at first.  The way we have it set now anybody can trigger a merge right?

tykeal:
IIRC right now we can't get remerges to work at all because they don't come in via the ghprb trigger which is the trigger that supports comment text scrapping.

jjb/edgex-templates-docker.yaml
```diff
@@ -0,0 +1,167 @@
...
+      - lf-infra-github-scm:
+          url: '{git-clone-url}{github-org}/{project}'
+          refspec: ''
+          branch: '{branch}'
+          submodule-recursive: '{submodule-recursive}'
+          choosing-strategy: default
+          jenkins-ssh-credential: '{jenkins-ssh-credential}'
+
+    triggers:
+      # no reason to add lf-infra-github-pr-trigger here since it doesn't
```
JPWKU, line 88: If you use the lf-infra-github-pr-trigger macro, it includes access settings like white-list and org list.  I'm assuming we will want to restrict the triggering of this job on PR's.

jjb/edgex-templates-docker.yaml
```diff
@@ -0,0 +1,167 @@
...
+      - lf-infra-github-scm:
+          url: '{git-clone-url}{github-org}/{project}'
+          refspec: '+refs/pull/*:refs/remotes/origin/pr/*'
+          branch: '{branch}'
+          submodule-recursive: '{submodule-recursive}'
+          choosing-strategy: default
+          jenkins-ssh-credential: '{jenkins-ssh-credential}'
+
+    triggers:
+      - github-pull-request:
```
JPWKU, line 67: This is available as a macro lf-infra-github-pr-trigger.

JPWKU (approved):
LGTM

zxiiro:
remerge

PR #23, opened 2017-08-10
Fixes Device SDK Tools JJB

trcox:
Project Generates source code for Device Service from template
Verify and Merge jobs fixed to generate, then run maven build on generated code

Proposed fix, welcome comments and changes

Signed-off-by: Tyler_Cox <Tyler_Cox@dell.com>

### DIFF
--- a/jjb/device/device-sdk-tools.yaml
+++ b/jjb/device/device-sdk-tools.yaml
@@ -5,10 +5,13 @@
     project-name: device-sdk-tools
     project: device-sdk-tools
     mvn-settings: device-sdk-tools-settings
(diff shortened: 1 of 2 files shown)


tykeal:
@trcox Things look good. Could you please squash the second change down into the first one though? We generally try to keep any final merge to be the discreet final form instead of an interative commit stream. Unless it actually makes sense to be a full patch series, but that's usually for larger changes so that it's more easily bisectable.

tykeal (commented):
This looks good

tykeal (changes requested):
Per my PR comment itself. Could you please squash this down into the other change?

trcox:
@tykeal Squashed!

tykeal (approved):

PR #29, opened 2017-08-31
Upgrade docker to latest version

JPWKU:
EdgeX wants docker version >=17.05 to
get access to multi-stage builds.

JIRA: RELENG-418

### DIFF
--- a/packer/provision/docker.sh
+++ b/packer/provision/docker.sh
@@ -7,9 +7,21 @@ set -xeu -o pipefail
 
 rh_changes() {
     echo "---> RH changes"
+    # Following directions from 
+    # https://docs.docker.com/engine/installation/linux/docker-ce/centos/
+
+    # remove old docker
+    yum remove -y docker docker-common docker-selinux docker-engine
+
+    # set up the repository
+    yum install -y yum-utils device-mapper-persistent-data lvm2
+    yum-config-manager --add-repo \
+    https://download.docker.com/linux/centos/docker-ce.repo
+    yum clean -y metadata
+
     # install docker and enable it
     echo "---> Installing docker"
-    yum install -y docker supervisor bridge-utils
+    yum install -y docker-ce supervisor bridge-utils
     systemctl enable docker
 
     # configure docker networking so that it does not conflict with LF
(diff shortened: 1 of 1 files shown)


packer/provision/docker.sh
```diff
@@ -7,9 +7,25 @@ set -xeu -o pipefail
...
+    # set up the repository
+    yum install -y yum-utils device-mapper-persistent-data lvm2
+    yum-config-manager --add-repo \
+    https://download.docker.com/linux/centos/docker-ce.repo
+
+    # update package repo
+    # figure out how to gpg verify the next step
+    # or can I just do yum -y update?
+
+    #####   sudo yum makecache fast
```
tykeal: If you force clear the metadata cache (yum clean metadata) and then follow that with the yum install -y it should take care of the GPG key IIRC.

Otherwise you will need to do the following:

rpm --import $GPG_KEY_URL

$GPG_KEY_URL will be in the repo file

packer/provision/docker.sh
```diff
@@ -20,7 +36,7 @@ DOCKER_NETWORK_OPTIONS='--bip=10.250.0.254/24'
 EOL
     # configure docker daemon to listen on port 5555 enabling remote
     # managment
-    sed -i -e "s#='--selinux-enabled'#='--selinux-enabled -H unix:///var/run/docker.sock -H tcp://0.0.0.0:5555'#g" /etc/sysconfig/docker
+    sed -i -e "s#='--selinux-enabled'#='--selinux-enabled -H unix:///var/run/docker.sock -H tcp://0.0.0.0:5555'#g" /etc/sysconfig/docker-ce
```
tykeal: Are you sure that the sysconfig file changes it's name to docker-ce?

JPWKU: I'm not sure...but `/etc/sysconfig/docker` doesn't exist

packer/provision/docker.sh
JPWKU: Yep that works.

packer/provision/docker.sh
JPWKU: Now docker recommends to put daemon config in `/etc/docker/daemon.yaml`.  The default location according to the docs.

JPWKU:
Closing this as it seems the DCO check is hung.

PR #41, opened 2017-09-19
Add daily release jobs for java and docker projects

JPWKU:

### DIFF
--- a/jjb/edgex-templates-docker.yaml
+++ b/jjb/edgex-templates-docker.yaml
@@ -166,3 +166,52 @@
           properties-file: 'env_inject.txt'
       # Execute any post build clean-up or testing
       - shell: '{post_build_script}'
+
+- job-template:
+    name: '{project-name}-{stream}-release-version-docker-daily-no-sonar'
+    # Job template for Docker daily release jobs
+    #
+    # The purpose of this job template is to run "docker build and docker push"
+    # for projects using this template.
+    #
+    # Required Variables:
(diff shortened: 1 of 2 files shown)


tykeal (changes requested):
In neither of these changes are you actually _building_  anything. That is, there's no builders section since the boiler_plate anchors don't do builders.

If you want to reuse a current builder section from one of the already defined jobs you will need to move that to a new anchor and modify the old job and the new jobs.

JPWKU:
@tykeal ah yes...cut/copy/paste got me all mixed up

JPWKU:
recheck

jjb/edgex-templates-docker.yaml
```diff
@@ -166,3 +166,52 @@
...
+    # Job template for Docker daily release jobs
+    #
+    # The purpose of this job template is to run "docker build and docker push"
+    # for projects using this template.
+    #
+    # Required Variables:
+    #     branch:    git branch (eg. stable/lithium or master)
+    <<: *docker_job_boiler_plate
+
+    docker_registry: '$DOCKER_REGISTRY:10004'
```
tykeal, line 181: Just a comment really, not a requested change (at this time) we should probably add some extra Jenkins ENV vars like DOCKER_STAGING_PORT DOCKER_SNAPSHOT_PORT and DOCKER_RELEASE_PORT so that these templates can be more easily moved to global-jjb and sort of "just work"

tykeal (approved):

PR #43, opened 2017-09-27
Use global jjb maven release job

JPWKU:
Have set trigger to build daily to
staging repo for device-bacnet.

### DIFF
--- a/jjb/defaults.yaml
+++ b/jjb/defaults.yaml
@@ -48,3 +48,7 @@
         -Djenkins
         -Dmaven.repo.local=/tmp/r
         -Dorg.ops4j.pax.url.mvn.localRepository=/tmp/r
+
+    # Nexus
+    staging-profile-id: a004df83a6249
+    mvn-staging-id: staging
--- a/jjb/device/device-bacnet.yaml
+++ b/jjb/device/device-bacnet.yaml
@@ -11,6 +11,7 @@
     archive-artifacts: ''
     artifact-version: '1.0.0-SNAPSHOT'
     build-node: centos7-docker-4c-2g
+    cron: 'H 11 * * *'
     stream:
       - 'master'
 
@@ -19,5 +20,5 @@
       - '{project-name}-{stream}-merge-java'
       - '{project-name}-{stream}-verify-docker'
       - '{project-name}-{stream}-merge-docker'
-      - '{project-name}-{stream}-release-version-java-daily-no-sonar'
+      - '{project-name}-github-maven-jobs'
       - '{project-name}-{stream}-release-version-docker-daily-no-sonar'


jjb/defaults.yaml
```diff
@@ -48,3 +48,6 @@
         -Djenkins
         -Dmaven.repo.local=/tmp/r
         -Dorg.ops4j.pax.url.mvn.localRepository=/tmp/r
+
+    # Nexus
+  staging-profile-id: a004df83a6249
```
zxiiro: Not sure if you mean to give every project the same staging-profile-id so thought I'd ask. In OpenDaylight this is a project level setting and every project has their own id.

jjb/device/device-bacnet.yaml
```diff
@@ -19,5 +19,7 @@
       - '{project-name}-{stream}-merge-java'
       - '{project-name}-{stream}-verify-docker'
       - '{project-name}-{stream}-merge-docker'
-      - '{project-name}-{stream}-release-version-java-daily-no-sonar'
+      - '{project-name} github-maven-jobs':
+          triggers:
+            - timed: 'H 11 * * *'
```
zxiiro: "triggers" is not a variable you have to use an actual variable here to override the trigger. Hmm now that I think about it I'm not sure how we can override for  github maybe we have to introduce a new variable.

and i'm not sure if it works to set something underneath a job template  (maybe it does, worth a try)

jjb/defaults.yaml
JPWKU: for now EdgeX is using the same profile...we will revisit after Barcelona release.

jjb/device/device-bacnet.yaml
JPWKU: I'm removing this now that we have it parameterized in global-jjb.

tykeal: @zxiiro on your comment about doing an override under a job template. I know that it works, Open-O used to and ONAP does use that pattern extensively.

tykeal:
@JPWKU looks like the change needs a rebase, plus two jobs that need to be re-triggered / run.

JPWKU:
@tykeal  rebased

JPWKU:
ok now to test in sandbox

jjb/defaults.yaml
```diff
@@ -48,3 +48,7 @@
         -Djenkins
         -Dmaven.repo.local=/tmp/r
         -Dorg.ops4j.pax.url.mvn.localRepository=/tmp/r
+
+    # Nexus
+    staging-profile-id: a004df83a6249
+    mvn-staging-id: staging
```
tykeal: Interestingly, you're missing the final newline at the end of the file and GitHub is calling you out on it.

JPWKU: heh fixing

tykeal (approved):

PR #45, opened 2017-09-29
Add reverse triggers for docker jobs

JPWKU:
This will serialize the flow artifacts
from java merge jobs to docker jobs.

JIRA: RELENG-468

### DIFF
--- a/jjb/edgex-templates-docker.yaml
+++ b/jjb/edgex-templates-docker.yaml
@@ -91,6 +91,8 @@
       - github
       - pollscm:
           cron: ''
+      - reverse:
+          jobs: '{project-name}-{stream}-merge-java'
 
 #################
 # JOB TEMPLATES #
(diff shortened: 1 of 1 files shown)


trcox:
@JPWKU Just so that I make sure that I am understanding this, the intent here is the following flow?

PR commits -> triggers verify-java and verify-docker checks, both of which must pass

PR accepted/merged -> triggers merge-java -> (upon completion) triggers verify-docker with new artifacts -> (upon completion) triggers merge-docker to update the snapshot Nexus3 Docker images?

JPWKU:
@trcox Basically yes...so we don't run into the situation where a java merge and a docker merge happen simultaneously.  What could happen there is the docker merge uses an older artifact than the java merge is building.

JPWKU:
I _think_ this takes care of that but I'm double checking nuances.

jjb/edgex-templates-docker.yaml
```diff
@@ -215,3 +217,5 @@
 
     triggers:
       - timed: 'H 11 * * *'
+      - reverse:
+          jobs: '{project-name}-{stream}-release-version-java-daily-no-sonar'
```
zxiiro: Should remove the timed here since java-daily is already timed.

JPWKU: Fixed.

zxiiro (approved):

PR #50, opened 2017-10-12
Fix docker release daily

JPWKU:
Updated to strip out -SNAPSHOT from dockerfiles
pom.


trcox (approved):

zxiiro (approved):
Seems fine to me but is it correct to be releasing the same version every day? versions are usually immutable.

I'll leave it to you to merge.

JPWKU:
@zxiiro the intent here is that the version is the same..I'm just producing a staging artifact for the docker container that uses the latest java staging artifact.

PR #57, opened 2017-11-14
Update branch to use GERRIT_BRANCH

JPWKU:
Also update refspec to use GERRIT_REFSPEC.
I've tested two of these and they work as intended and do the right branch build.

### DIFF
--- a/jjb/edgex-templates-java.yaml
+++ b/jjb/edgex-templates-java.yaml
@@ -27,7 +27,7 @@
     parameters:
       - lf-infra-parameters:
           project: '{project}'
-          branch: '{branch}'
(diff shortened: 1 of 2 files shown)


JPWKU:
recheck

jjb/edgex-templates-java.yaml
```diff
@@ -27,7 +27,7 @@
     parameters:
       - lf-infra-parameters:
           project: '{project}'
-          branch: '{branch}'
+          branch: '$GERRIT_BRANCH'
```
trcox, line 30: Missing refspec reference here?

PR #63, opened 2017-11-20
Add new device jobs

JPWKU:
Add jobs for newly created
device-controller and
device-domain

### DIFF
--- a/jjb/device/device-controller.yaml
+++ b/jjb/device/device-controller.yaml
@@ -1,0 +1,17 @@
+---
+
+- project:
+    name: device-controller
+    project-name: device-controller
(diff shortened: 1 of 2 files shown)


trcox:
@JPWKU These will be new device domain libraries, templating should look like [core-domain](https://github.com/edgexfoundry/ci-management/blob/master/jjb/core/core-domain.yaml), i.e. no Docker builds.

trcox (changes requested):
@JPWKU These will be new device domain libraries, templating should look like core-domain, i.e. no Docker builds.

trcox (approved):

PR #68, opened 2017-11-28
Fix docker tagging

JPWKU:
Follow the maven convention for docker
tagging based on the stream and corresponding
version in pom.


trcox (dismissed):

JPWKU:
@trcox I was testing in the sandbox and noticed the xmlstarlet command was wrong...testing a fix now.

jwagantall (approved):

PR #92, opened 2018-03-15
Upgrade global-jjb to v0.17.0

zxiiro:
Pull in the latest global-jjb v0.17.0 and ensure that we configure the global-variables files so that the new jenkins-cfg job can manage it.


JPWKU (approved):
LGTM

PR #95, opened 2018-03-20
Common Packer builder

zxiiro:
Setup the Common Packer builder image type.


zxiiro:
Packer Verify is expected to fail. Trying to fix that. Merging this patch as we need the new builder type to get the dependencies we need in the image.

PR #96, opened 2018-03-20
Common Packer docker

zxiiro:
Pull in common packer docker image.


zxiiro:
The Packer Verify job is expected to fail. Merging as we're working on getting it fixed.

zxiiro:
remerge

PR #97, opened 2018-03-20
Remove basebuild as replaced by builder

zxiiro:
Builder replaces basebuild so remove it.

Signed-off-by: Thanh Ha <thanh.ha@linuxfoundation.org>


zxiiro:
Merging as we are working on resolving the packer verify issues.

PR #107, opened 2018-04-18
Use image name that is in edgexci cloud

JPWKU:
I tried to modify that name but the
edgexci user did not have permissions
to do so.


tykeal:
recheck

tykeal (approved):

PR #114, opened 2018-05-04
Publish docker images to nexus and dockerhub

JPWKU:

### DIFF
--- a/jjb/edgex-go/edgex-go.yaml
+++ b/jjb/edgex-go/edgex-go.yaml
@@ -4,18 +4,22 @@
     project-name: edgex-go
     project: edgex-go
     mvn-settings: edgex-go-settings
+    pre_build_script: 'make prepare && make test'
+    build_script: 'make build && make docker'
     stream:
       - 'master':
           branch: 'master'
 
     jobs:
       - '{project-name}-{stream}-verify-go':
-          pre_build_script: 'make prepare && make test'
-          build_script: 'make build && make docker'
           status-context: '{project-name}-verify'
       - '{project-name}-{stream}-verify-go-arm':
-          pre_build_script: 'make prepare && make test'
-          build_script: 'make build && make docker'
           status-context: '{project-name}-verify-arm'
-      - '{project-name}-{stream}-merge-go'
-      - '{project-name}-{stream}-merge-go-arm'
+      - '{project-name}-{stream}-merge-go':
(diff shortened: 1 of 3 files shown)


tsconn23:
@JPWKU What is "JJB"? I see the new files in this PR in a jjb directory and I notice that the specific verification step has failed. Is that a concern?

JPWKU:
@tsconn23 JJB is Jenkins Job Builder, which is the tool we use to provision all of our Jenkins jobs.  This is the JJB config for publishing artifacts finally.  There are some bugs still to work out, a failed verify on JJB means I have something wrong in it.

JPWKU:
Hmm I just realized we will have issues with this script on Caviums static ARM build machine, but we can go ahead and merge and debug from there.

JPWKU:
@feclare The issue I see with this script is that we can not trust that only a specific build's docker images are the only edgexfoundry docker images on the static build machine.  To get around this I think we can modify the Makefile to interpolate the Jenkins build number into a tag that is generated.  After that I can modify this script to only query for that tag with a specific build number.  Let me think on it a bit.

feclare:
@JPWKU The board used is dedicated to the edgex jenkins jobs. There should not be docker images not generated by edgex.
But I understand the trust issue, anything we can do to help just ask.

JPWKU:
@feclare I submitted a patch on edgex-go that modifies the makefile and gives me specific docker tags I can query for.  I think this will solve the issue.  Consider that two patches are both running builds simultaneously, how can I differentiate the docker images generated?  I think tagging with the SHA of the build solves that issues.

JPWKU:
@feclare as a matter of fact I think we can remove the other tags, but let me get this sorted before I do that.

tsconn23:
@JPWKU "but let me get this sorted before I do that."

Does that mean you want us to wait on merging this PR or is it ready?

JPWKU:
@tsconn23 I have a bit more coming, I'll let you know when it's ready.

JPWKU:
@tsconn23 I think this is ready for merge now.  There are likely still things to work out but we won't know until it starts running.  I'll keep an eye on it after merge.

JPWKU:
We will likely have some race condition on docker tag an docker push on Caviums static machine.  I'm still figuring how to work around that.

tsconn23 (approved):

PR #123, opened 2018-05-11
Forcibly remove docker images after each build

JPWKU:
This is to accomodate a static build machine.

### DIFF
--- a/jjb/edgex-macros.yaml
+++ b/jjb/edgex-macros.yaml
@@ -7,3 +7,13 @@
           recipients: EdgeX-Jenkins-Alerts@lists.edgexfoundry.org
           notify-every-unstable-build: true
           send-to-individuals: true
+
+- builder:
+    name: edgex-provide-docker-cleanup
+    builders:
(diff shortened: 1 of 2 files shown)


JPWKU:
@feclare This macro will only run on jobs executing on the cavium machine.

jjb/edgex-macros.yaml
```diff
@@ -1,17 +1,9 @@
...
-- property:
-    name: 'edgex-go-amd64'
-    properties:
-      - inject:
-          properties-content: |
-            GOPATH=$WORKSPACE
-            GOARCH=amd64
-
+- builder:
+    name: lf-provide-docker-cleanup
```
tykeal: We should not be adding things to the lf- macro / template namespace inside projects. That is reserved for global-jjb use only. Instead, if you want to be safe and namespace this, use edgex-

tykeal (changes requested):
Please fix the namespace on this macro

tykeal (approved):

PR #128, opened 2018-05-16
Build snapcraft ubuntu image

JPWKU:

### DIFF
--- a/jjb/ci-management/packer.yaml
+++ b/jjb/ci-management/packer.yaml
@@ -36,3 +36,21 @@
       - centos-7
       - ubuntu-16.04
     templates: docker
+
+- project:
+    name: packer-snapcraft-jobs
+    jobs:
+      - github-packer-merge
+
+    project: ci-management
+    project-name: ci-management
+    branch: master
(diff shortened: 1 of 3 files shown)


packer/templates/ubuntu.json
```diff
@@ -0,0 +1,64 @@
...
+    "cloud_network": null,
+    "cloud_tenant": null,
+    "cloud_user_data": null,
+    "ssh_user": null,
+    "ssh_proxy_host": ""
+  },
+  "builders": [
+    {
+      "name": "vexxhost",
+      "image_name": "ZZCI - {{user `distro`}} - snapcraft - {{isotime \"20060102-1504\"}}",
```
tykeal: You're you're naming the built image with an identifier of snapcraft but the file is named ubuntu.json please change the file name to match the image identifier.

tykeal (changes requested):
packer/templates/ubuntu.json does not match with the as built image name. Please fix the file name to match the as built image identifier (snapcraft). Both of those keys need to match.

tykeal (dismissed):

PR #144, opened 2018-06-25
Add snap macroon provider and cleanup

JPWKU:
This will enable non-interactive login to
the Ubuntu Snapcraft store.

### DIFF
--- a/jjb/edgex-macros.yaml
+++ b/jjb/edgex-macros.yaml
@@ -34,3 +34,18 @@
       - shell: !include-raw-escape:
           - global-jjb/shell/lftools-install.sh
           - shell/edgex-publish-docs.sh
+
+- builder:
+    name: edgex-provide-snap-cleanup
+    builders:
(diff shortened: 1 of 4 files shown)


jjb/edgex-macros.yaml
```diff
@@ -34,3 +34,25 @@
...
+    name: edgex-provide-snap-cleanup
+    builders:
+      - shell: |
+          #!/bin/bash
+          set +e  # DO NOT cause build failure if any of the rm calls fail.
+
+          rm -f "$HOME/EdgeX"
+
+          # DO NOT fail build if any of the above lines fail.
+          exit 0
```
tykeal: You shouldn't be inlining shell scripts. Doing so makes it impossible to run ShellCheck against the code.

tykeal (changes requested):
Inlining shell scripts should in general be avoided. This one isn't doing too much, but it would still be best to move this out to a separate script so that ShellCheck can actually validate the code

tykeal (approved):

PR #149, opened 2018-06-28
Reconfigure to publish and release built snap

JPWKU:

### DIFF
--- a/shell/edgex-publish-snap.sh
+++ b/shell/edgex-publish-snap.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+# Ensure we fail the job if any steps fail
+set -e -o pipefail
+snapcraft login --with $HOME/EdgeX
+
+# Build snap and grab the generated snap name
+SNAP_TO_PUSH=$(snapcraft | awk '/Snapped/ {print $2}')
+
+# Push the generated snap and grab the revision number
+REVISION=$(snapcraft push $SNAP_TO_PUSH | awk '/Revision/ {print $2}')
+
+# Release the snap to channel
+snapcraft release edgexfoundry $REVISION --channel=$SNAP_CHANNEL
--- a/jjb/edgex-go/edgex-go-snap.yaml
+++ b/jjb/edgex-go/edgex-go-snap.yaml
@@ -4,14 +4,14 @@
     project-name: edgex-go-snap
     project: edgex-go
     mvn-settings: edgex-go-settings
-    pre_build_script: snapcraft login --with "$HOME/EdgeX"
-    build_script: 'snapcraft'
-    post_build_script: ''
+    build-script: !include-raw-escape: shell/edgex-publish-snap.sh
     stream:
       - 'master':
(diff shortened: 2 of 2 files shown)


shell/edgex-publish-snap.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+# Ensure we fail the job if any steps fail
+set -e -o pipefail
+snapcraft login --with $HOME/EdgeX
+
+# Build snap and grab the generated snap name
+SNAP_TO_PUSH=$(snapcraft | awk '/Snapped/ {print $2}')
```
anonymouse64, line 7: This line seems a little iffy to me, while the phrase `Snapped` isn't likely to show up in the build output there is a ton of stuff printed out during the build. I've found while automating to just grab the name of the `*.snap` file to be easier like this:
```
snapcraft push edgexfoundry*.snap
```
The only problem you might run into with this is if you end up building multiple architectures in the same workspace, but then you can filter with the architecture:
```
snapcraft push edgexfoundry*_amd64.snap
```

If you really want to keep with the current method I'd suggest making the string matching more precise a la something like
```
SNAP_TO_PUSH=$(snapcraft | grep -oP 'Snapped \K(edgexfoundry_.+\.snap)')
```

(and you could even add in the architecture if you really wanted to be careful)

shell/edgex-publish-snap.sh
```diff
@@ -0,0 +1,13 @@
...
+
+# Push the generated snap and grab the revision number
+REVISION=$(snapcraft push $SNAP_TO_PUSH | awk '/Revision/ {print $2}')
+
+# Release the snap to channel
+snapcraft release edgexfoundry $REVISION --channel=$SNAP_CHANNEL
```
anonymouse64, line 13: You don't need the `--channel=` here, you can just use
```
snapcraft release edgexfoundry $REVISION $SNAP_CHANNEL
```

(at least that's with the most recent version of snapcraft: `snapcraft, version 2.42.1+git25.25043ab`)

anonymouse64:
@JPWKU I appear to be a little late to the game on this one, but I would still like to see the changes I mentioned above merged, so if you want to open a new PR with those changes I would appreciate it. Thanks!

JPWKU:
@anonymouse64 Those are good suggestions, I'll PR them and tag you shortly.

shell/edgex-publish-snap.sh
JPWKU: Ok I think we will be good with `snapcraft push edgexfoundry*.snap` as we will not be building in the same workspace (dynamic build machines)

shell/edgex-publish-snap.sh
JPWKU: Hmm I'm not sure which version we are on yet, will likely keep this unless it doesn't work.

anonymouse64: Sure, that's fine just at some future point this may not be supported as an option, not sure

PR #150, opened 2018-06-28
Address comments from Ian

JPWKU:
Also added arm64 node build


anonymouse64 (approved):
Looks like this will also enable the arm64 building, which I can't really review but looks good to me anyways.

PR #154, opened 2018-06-28
[DO NOT MERGE] Add arm64 Ubuntu images for docker and snapcraft

JPWKU:


tykeal (dismissed):
We should update comman-packer so that we don't have a proliferation of templates just because of different architectures.

JPWKU:
Doing this in common packer instead.

PR #204, opened 2018-08-16
shell/edgex-publish-snap.sh: remove incorrect --channel option

### LINKED ISSUE #203: snap california build fails to release
See for example : https://jenkins.edgexfoundry.org/view/Snap/job/edgex-go-snap-california-stage-snap/54/consoleFull

After having pushed the snap up, it fails to release here:

```
Processing...|                                                                 
22:16:38 Error: no such option: --channel
22:16:38 Build step 'Execute shell' marked build as failure
```
I think this is because in the snap stage job here: https://github.com/edgexfoundry/ci-management/blob/master/shell/edgex-publish-snap.sh#L14

we have

```
snapcraft release edgexfoundry $REVISION --channel=$SNAP_CHANNEL
```
which should be :
```
snapcraft release edgexfoundry $REVISION $SNAP_CHANNEL
```

anonymouse64:
snapcraft just takes the channel as an argument, not as a --channel option.

Fixes #203 

Should fix builds that succeed in building the snap, but fail to release the snap such as https://jenkins.edgexfoundry.org/view/Snap/job/edgex-go-snap-california-stage-snap/54/consoleFull

### DIFF
--- a/shell/edgex-publish-snap.sh
+++ b/shell/edgex-publish-snap.sh
@@ -11,4 +11,4 @@ snapcraft
 REVISION=$(snapcraft push edgexfoundry*.snap | awk '/Revision/ {print $2}')
 
 # Release the snap to channel
-snapcraft release edgexfoundry $REVISION --channel=$SNAP_CHANNEL
+snapcraft release edgexfoundry $REVISION $SNAP_CHANNEL


JPWKU (approved):
LGTM

PR #211, opened 2018-09-22
Enable snap docker builds and remove ubuntu node requirements

anonymouse64:
This PR enables using the build script for the snap that uses docker to build the snap. Because we're now using docker, we don't need to use specific OS requirements, and we can simplify the job definitions.
Additionally, renaming the PR comment to just '{project-name}-{stream}-verify', as the project-name already specifies that it's for the snap, and on GitHub for example the comment shows up as "edgex-go-snap-snap-master-verfiy", etc.
Lastly, because the build script + docker images now contain the logic necessary to publish the snap using the JENKINS_URL and JOB_NAME environment variables, we can delete the shell/edgex-publish-snap.sh script here as well.

After discussion in the devops meeting on October 1st 2018, we would like to merge this PR first, then we can see if the snap tests in edgex-go pass from this PR https://github.com/edgexfoundry/edgex-go/pull/563.


JPWKU:
@anonymouse64 Checking this patch out.

JPWKU:
Ok this looks good to me, going to checkout the the edgex-go related PR.

anonymouse64:
@jwagantall and @tykeal can y'all take a look at this PR (and the corresponding one for the edgex-go repo linked to in the PR description?

tykeal (dismissed):

tykeal:
@anonymouse64 This needs to be rebased onto the current master before it can be merged, but it looks good to me.

anonymouse64:
@tykeal thanks for reviewing, I just rebased the branch.

tsconn23 (approved):

PR #220, opened 2018-10-08
edgex-go-snap.yaml: specify docker nodes for snap jobs

anonymouse64:
We actually do need to specify the build nodes for the snap build jobs, as they require docker, and only these hosts have docker available.


JPWKU (dismissed):
LGTM

JPWKU (approved):
LGTM

PR #224, opened 2018-10-09
Migrate from cavium arm64

JPWKU:


tsconn23 (dismissed):
SHIP IT!!!

tsconn23 (approved):

jjb/edgex-go/edgex-go-snap.yaml
```diff
@@ -14,12 +14,13 @@
...
      - '{project-name}-{stream}-stage-snap-arm':
-         build-node: cavium-arm64
+         build-node: ubuntu18.04-docker-arm64-4c-2g
      - '{project-name}-{stream}-stage-snap':
          build-node: centos7-docker-4c-2g
      - '{project-name}-{stream}-verify-snap-arm':
-         build-node: cavium-arm64
-         status-context: '{project-name}-{stream}-verify-arm'
+         build-node: ubuntu18.04-docker-arm64-4c-2g
+         build_script: 'snapcraft clean && snapcraft'
```
anonymouse64, line 22: This is wrong, we shouldn't be running `snapcraft` directly here anymore it will fail on Ubuntu 18.04. We should use the new `build.sh` script which is the default for the snap jobs here. I think if you just delete this line it should fix things.

JPWKU: Ok, I had a merge conflict here and apparently kept some old cruft.

PR #228, opened 2018-10-12
Install second golang in opt

JPWKU:
This installs golang 1.10.0 in /opt/golang/110/go/bin.
The user will need to specify this in their PATH or alias go to
the above full path.  Alternatively they can set GOROOT.


jpwhitemn (approved):
We have to try it for security services.  I need to admonish all groups not to move to alternate version of Go or other infrastructure without consultation of the community.

PR #238, opened 2018-10-18
Add facility to use a different version of golang

JPWKU:
This patch adds GOROOT with a default.  The user can set it to
any arbitray go installation.


JPWKU:
Testing this real quick in sandbox

JPWKU:
Ok looks good in sandbox

tykeal (commented):
Why does this keep setting / resetting the PATH? It seems rather weird

JPWKU:
It's a normal step in setting up a golang install.

tykeal (approved):
That PATH resetting still seems super weird to me, but whatever.

PR #243, opened 2018-10-30
Add jobs for edgex-ui-go

JPWKU:

### DIFF
--- a/shell/edgexfoundry-edgex-ui-go-push.sh
+++ b/shell/edgexfoundry-edgex-ui-go-push.sh
@@ -1,0 +1,93 @@
+#!/bin/bash
+# Ensure we fail the job if any steps fail
+set -e -o pipefail
+
+# Required parameters:
+#
+#  VERSION: Set this value based on the VERSION file in the project root.
+#  DEPLOY_TYPE: Can be `snapshot`, `staging` or`release`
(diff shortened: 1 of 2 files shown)


shell/edgexfoundry-edgex-ui-go-push.sh
```diff
@@ -0,0 +1,93 @@
...
+#    `snapshot` will push docker images to:
+#       1) nexus3.edgexfoundry.org:10003 with a GIT_SHA-VERSION tag
+#    `staging` will push docker images to:
+#       1) nexus3.edgexfoundry.org:10004 with the `latest` tag
+#       2) edgexfoundry dockerhub with the `latest` tag
+#    `release` will push docker images to:
+#       1) nexus3.edgexfoundry.org:10002 with the `latest` tag and `VERSION` tag
+#       2) edgexfoundry dockerhub with the `latest` tag and `VERSION` tag
+#
+if [[ `uname -m` != "x86_64" ]]; then
```
tykeal: instead of using backtick shell outs you should use $() shell outs

So instead of `uname -m` you should use $(uname -m)

JPWKU: Fixed.

tykeal (approved):

PR #266, opened 2019-01-08
Us2961/ci device sdk c

jamesrgregg:
- add device-sdk-c jjb template
@JPWKU please review


jjb/device/device-sdk-c.yaml
```diff
@@ -3,16 +3,15 @@
...
       - 'master':
-          branch: 'master'
-      - 'edinburgh':
-          branch: 'edinburgh'            
+          branch: 'master'          
     jobs:
-      - '{project-name}-cmake-stage-{stream}'
-      - '{project-name}-cmake-verify-{stream}'
\ No newline at end of file
+      - '{project-name}-cmake-verify-{stream}'
```
JPWKU, line 16: Because this job name string is shared for two templates (one for Github and one for Gerrit) we will need to use the id.
So github-cmake-verify and github-cmake-stage

Though, I'm not sure we will need stage..since this is an sdk.  Will there be artifacts staged?

jamesrgregg: the only artifact is in the following:  'device-sdk-c/build/release/*.gz'
specifically named csdk-1.0.0.tar.gz

jamesrgregg: @JPWKU why do we reference github-cmake ?  Is that just an example project name you're referring to?

jamesrgregg: I think I I figured out what you were referring to - tried it with {project-name} and without - using the id (github-cmake-verify & github-cmake-stage).  No luck.

PR #267, opened 2019-01-09
Jamesrgregg us2961/ci device sdk c

jamesrgregg:
- set job to use id


jamesrgregg:
@JPWKU No luck tonight getting this to build.  I basically hard coded all of the build parameters I thought it would need.  The build logs are not helpful because the failure reason isn't clear at the end.  I don't even see it cloning the source code and yet I set the git_url: in the jjb template.  Perhaps if you have time we can take a closer look at this together.  Thanks for your review.

jjb/device/device-sdk-c.yaml
```diff
@@ -0,0 +1,25 @@
...
+    git_url: '$GIT_URL/$PROJECT'
+    build_script: '$WORKSPACE/device-sdk-c/scripts/build.sh'
+    build_dir: '$WORKSPACE/device-sdk-c/scripts/build'
+    cron: 'H 11 * * *'
+    stream:
+      - 'master':
+          branch: 'master'          
+    jobs:
+      - '{project-name}-github-cmake-verify'
+      - '{project-name}-github-cmake-stage'
```
zxiiro: The jobs should be:

```
jobs:
  - github-cmake-verify
  - github-cmake-stage
```
Remove the "{project-name}-" part of the name.

You can see a list of  global-jjb job template projects here:

https://docs.releng.linuxfoundation.org/projects/global-jjb/en/latest/jjb/lf-c-cpp-jobs.html

I wouldn't use the {project-name}-cmake-verify-{stream}"  version of the name because it's a duplicate template in JJB and would cause JJB to be confused about if it should pull gerrit or github one.

Hope this helps

zxiiro (changes requested):

jjb/device/device-sdk-c.yaml
jamesrgregg: @zxiiro Thank you for the reply.  I was confused by the same document as the error I was getting when deploying in the Jenkins sandbox, was referencing something about not being able to find the suitable template.  The document referenced the template with {project-name}-...{stream}.  Thanks for the clarification. I have made the change as suggested and it still looks like JJB Verify is not good.

zxiiro: @jamesrgregg  the failure this time is "JenkinsJobsException: nexus-group-id parameter missing to format {nexus-group-id}" this means one of the templates (the stage template) you are pulling in is expecting to be passed parameter called "nexus-group-id" and but it was not provided in the project definition.

nexus-group-id is typically in a fully qualified domain kind of definition so (I'm guessing here) but it should be "org.edgexfoundry.device-sdk-c" Assuming (device-sdk-c) is the name of the project (I got it from the definition in the proposed file. @JPWKU  should be able to confirm.

Jenkins Job Builder (JJB)'s log system is a bit difficult to navigate for errors. Typically what I do is jump to the bottom of the console logs and scroll up usually should be able to spot the error this way. If not doing a search for "Exception" or "Error" sometimes helps.

Hope this helps.

jamesrgregg: @zxiiro I will make that change.  thank you for your help and excellent suggestion for finding the error / exceptions easier in the build log.

jamesrgregg:
@zxiiro All checks have passed which is good, but I don't have a high level of confidence that the job will do what it needs to do.. execute the ./scripts/build.sh and create the sdk tar.gz file.  I"m not seeing that it goes to the project source and even pulls any code.  Is there something still missing, or is there more to this once it gets merged?  
I did try to use a fqdn for the nexus-group-id, but something didn't go right with one of the docker-verify jobs, so I backed it out.
I will work on squash and rebase next, but would appreciate a bit more clarification please.

zxiiro:
Hi @jamesrgregg for some background since I wrote these 2 jobs, so I originally wrote these 2 job templates for the ASWF as they are essentially all c/c++ library projects. However the project is very new and we haven't fully settled on if these job templates are the best way to build cmake projects, we are still in active discussion on how we want to manage c/c++ libraries and there were some discussions regarding using something like conan.io but this is all in the investigative phase at the moment and nothing set in stone yet.

With that said the 2 jobs "cmake-verify" and "cmake-stage" are essentially the same job with one exception (which I'll discuss a bit later). The jobs are designed with the following build phases.

0. Fetch git repo
1. Run {pre-build} 

Essentially pre-build is an prep step allowing you to run a bash shell to preform any necessary preparation steps before the build is run such as fetching dependencies that might not exist in the VMs or massaging data to set the version pretty much anything that bash will let you do.

Some examples: 

https://github.com/AcademySoftwareFoundation/ci-management/blob/master/jjb/opencolorio/opencolorio.yaml#L16-L21

https://github.com/AcademySoftwareFoundation/ci-management/blob/master/jjb/openvdb/openvdb.yaml#L21-L30

2. Run cmake-build.sh script (https://github.com/lfit/releng-global-jjb/blob/master/shell/cmake-build.sh)

Essentially what this script does 2 things:

2.a. Runs `cmake -DCMAKE_INSTALL_PREFIX="$INSTALL_PREFIX" $cmake_opts` 
where $INSTALL_PREFIX defaults to /usr/local/bin but can be overrided and $cmake_opts defaults to a empty string but the job user can pass additional parameters specific to their project.

2.b. Creates a tar.xz file of the build result and stores it in "$WORKSPACE/dist/$project.tar.gz where "$project" is the "project-name" parameter passed in the JJB YAML template.

At this point this is the end of the purpose of the "cmake-verify" job and it will report the result succes|fail back to GitHub.

3. (Only applies to cmake-stage) Runs the cmake-stage.sh script (https://github.com/lfit/releng-global-jjb/blob/master/shell/cmake-stage.sh)

In this step it uses Sonatype Nexus 2 service to prepare a staging repository containing the tar.xz file produced in step 2 to prepare for a release. This staging repo can be used by the project to test and validate the release. Once validated the staging repo can be moved to a final release area in Sonatype Nexus 2 and made available to the public.

Admittedly this staging process is a very Java / Maven workflow and we are shoving a C/C++ project into a Java / Maven workflow here and may be a downside to this method. This is a discussion we are still trying to figure out and do not have a best practice at the moment on what to do although conan.io does sound very interesting to the ASWF project.

I hope this background and explanation is helpful to you in deciding if this is the right path for you or if you need to design a new workflow for your project.

Also I am happy to help answer questions on IRC a bunch of us CI folks hang out on #lf-releng @ irc.freenode.net in case you want to jump on an have a chat.

PR #268, opened 2019-01-09
Add device-grove-c jobs

lranjbar:

### DIFF
--- a/jjb/device/device-grove-c.yaml
+++ b/jjb/device/device-grove-c.yaml
@@ -21,7 +21,9 @@
           docker_build_args: '-f scripts/Dockerfile.alpine-3.8'
           docker_name: docker-device-grove-c-arm64
           status-context: '{project-name}-{stream}-docker-verify-arm'
+          post_build_script: !include-raw-escape: shell/edgexfoundry-device-grove-c-docker-push.sh
       - '{project-name}-{stream}-merge-docker-arm':
           build-node: ubuntu18.04-docker-arm64-4c-2g
           docker_build_args: '-f scripts/Dockerfile.alpine-3.8'
           docker_name: docker-device-grove-c-arm64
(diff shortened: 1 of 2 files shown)


jamesrgregg:
@JPWKU @SteveOss we need a reviewer on this one please.

SteveOss (commented):
LGTM, but am not an expert on this stuff.

JPWKU:
@lranjbar I think you mentioned this yesterday; Is this only being built for arm architecture?

SteveOss:
@JPWKU To use will need to be built on appropriate system/architecture. But for a CI check, a compile on a single system would probably suffice.

lranjbar:
@JPWKU From what I can tell in the repo - this code is targeting Raspberry PI 3+ (ARM) only. I wasn't sure how this would change the jobs needed in the CI pipeline. This was my first guess.

jamesrgregg (commented):
Looks like we need to define where the image will go after build.

jpwhitemn (dismissed):
LGTM

lranjbar:
@jpwhitemn Please review I added the docker push shell script like the other jobs have. thanks!

jamesrgregg:
@MightyNerdEric can you please take a look at this one and approve if all looks good.  Thank you.

eb-oss (changes requested):
Ok, after adding configuration to the Jenkins machines and running some tests, it looks like these jobs are expecting a pom.xml file in the project root. Are there other C builds that use the {verify,merge}-docker-arm jobs? If not, these should be thoroughly tested in the Jenkins sandbox to ensure that they are working as needed.

lranjbar:
@MightyNerdEric Thanks for looking into this! From what I can tell in EdgeX foundry there hasn't been C builds before these. I thought this template might be applicable but turns out it isn't as generic as I thought... back to the drawing board.

PR #270, opened 2019-01-14
Jrgregg/add cmake

jamesrgregg:


JPWKU:
Ok, now we will wait for verify, then we can create the job `ci-managment-packer-merge-centos-7-docker` and `ci-management-packer-merge-ubuntu-18.04-arm64` in the sandbox, these will run fully in the sandbox.

eb-oss (approved):

PR #277, opened 2019-02-01
Using the EdgeX docker template to build device-sdk-c

ernestojeda:
We are using the `jjb/edgex-templates-docker.yaml` template for the device-sdk-c build due to the fact that the developers are using a Docker image to build the C code. When run, the docker image that is generated, actually compiles the code and generates the deliverable, a tar.gz file in a release directory. So a post build action (shell/device_sdk_c_build.sh) is required to run a container and with the Jenkins workspace bind mounted in to be able to copy the tar.gz out of the container. The docker image in this repo is not really needed in the end and can probably be discarded. Other techniques could potentially be used to extract the tar file including modify the Dockerfile in the device-sdk-c repo to use a multi-stage build compile the c code in the first stage, then in the final stage, copy the deliverable to a scratch image.

### DIFF
--- a/jjb/device/device-sdk-c.yaml
+++ b/jjb/device/device-sdk-c.yaml
@@ -1,20 +1,27 @@
 ---
+
 - project:
     name: device-sdk-c
     project-name: device-sdk-c
     project: device-sdk-c
-    build-node: centos7-builder-2c-1g
-    nexus-group-id: device-sdk-c
     mvn-settings: device-sdk-c-settings
-    archive-artifacts: '$WORKSPACE/device-sdk-c/build/release/*.gz'
-    pre_build_script: ''
-    git_url: '$GIT_URL/$PROJECT'
-    build_script: '$WORKSPACE/device-sdk-c/scripts/build.sh'
-    build_dir: '$WORKSPACE/device-sdk-c/scripts/build'
+    docker_name: docker-device-sdk-c
+    docker_root: ''
+    docker_build_args: '-f scripts/Dockerfile.alpine-3.7'
+    archive-artifacts: '**/release/*.tar.gz'
+    post_build_script: !include-raw-escape: shell/device_sdk_c_build.sh
+    build-node: centos7-docker-4c-2g
     cron: 'H 11 * * *'
     stream:
       - 'master':
-          branch: 'master'          
+          branch: 'master'
+          docker_tag: 'master'
     jobs:
-      - 'github-cmake-verify'
-      - 'github-cmake-stage'
+      - '{project-name}-{stream}-verify-docker'
+      - '{project-name}-{stream}-merge-docker'
+      - '{project-name}-{stream}-verify-docker-arm':
+          build-node: ubuntu18.04-docker-arm64-4c-2g
+          docker_name: docker-device-sdk-c-arm64
+      - '{project-name}-{stream}-merge-docker-arm':
+          build-node: ubuntu18.04-docker-arm64-4c-2g
+          docker_name: docker-device-sdk-c-arm64
--- a/shell/device_sdk_c_build.sh
+++ b/shell/device_sdk_c_build.sh
@@ -1,0 +1,19 @@
+#!/bin/bash
+set -ex -o pipefail
+
+echo "--> device_sdk_c_build.sh"
+
+userId=`id -u`
+groupId=`id -g`
+
+mkdir $WORKSPACE/release
+
+DOCKER_BUILD_CONTAINER=${DOCKER_NAME}-builder-${BUILD_ID}
+
+echo "Starting build container and generating tar.gz"
+# --privileged is required because for some reason when docker bind mounts the
+# on the workspace on the Jenkins build server, it is Read-Only
+docker run --rm -e UID=$userId -e GID=$groupId --privileged -v $WORKSPACE/release:/edgex-c-sdk/build/release ${DOCKER_IMAGE}
+
+echo "Contents of release directory:"
+ls -al $WORKSPACE/release


ernestojeda:
@MightyNerdEric @jamesrgregg Please add reviewers as needed.

jjb/device/device-sdk-c.yaml
```diff
@@ -1,20 +1,28 @@
...
-    pre_build_script: ''
-    git_url: '$GIT_URL/$PROJECT'
-    build_script: '$WORKSPACE/device-sdk-c/scripts/build.sh'
-    build_dir: '$WORKSPACE/device-sdk-c/scripts/build'
+    docker_name: docker-device-sdk-c
+    docker_root: ''
+    docker_build_args: '-f scripts/Dockerfile.alpine-3.7'
+    archive-artifacts: '**/release/*.tar.gz'
+    post_build_script: !include-raw-escape: shell/device_sdk_c_build.sh
+    artifact-version: '0.1.0-SNAPSHOT'
```
ernestojeda: Not really sure how the `artifact-version` is used here. I think this was a copy/paste from another template.

jamesrgregg: It was a copy/paste of the template based on device-grove-c. 
The build script is tagging the artifact with 1.0.0 so it may not be needed.

jamesrgregg (commented):
LGTM - all checks pass

jjb/device/device-sdk-c.yaml
eb-oss: Have you tested the job without this line? If it's not needed, it should be removed. And even if it is needed, you should probably make it something more appropriate (that is, a generic tag that indicates that no version was passed to the job).

jjb/device/device-sdk-c.yaml
```diff
@@ -1,20 +1,28 @@
 ---
+
 - project:
     name: device-sdk-c
-    project-name: device-sdk-c
+    project-name:  device-sdk-c
```
eb-oss: Please remove this additional space.

shell/device_sdk_c_build.sh
```diff
@@ -0,0 +1,19 @@
...
+GID=`id -g`
+
+mkdir $WORKSPACE/release
+
+DOCKER_BUILD_CONTAINER=${DOCKER_NAME}-builder-${BUILD_ID}
+
+echo "Starting build container and generating tar.gz"
+# --privileged is required because for some reason when docker bind mounts the
+# on the workspace on the Jenkins build server, it is Read-Only
+docker run --rm -e UID -e GID --privileged -v $WORKSPACE/release:/edgex-c-sdk/build/release ${DOCKER_IMAGE}
```
eb-oss: Should `UID` and `GID` be `$UID` and `$GID`?

jjb/device/device-sdk-c.yaml
```diff
@@ -1,20 +1,28 @@
...
-    archive-artifacts: '$WORKSPACE/device-sdk-c/build/release/*.gz'
-    pre_build_script: ''
-    git_url: '$GIT_URL/$PROJECT'
-    build_script: '$WORKSPACE/device-sdk-c/scripts/build.sh'
-    build_dir: '$WORKSPACE/device-sdk-c/scripts/build'
+    docker_name: docker-device-sdk-c
+    docker_root: ''
+    docker_build_args: '-f scripts/Dockerfile.alpine-3.7'
+    archive-artifacts: '**/release/*.tar.gz'
+    post_build_script: !include-raw-escape: shell/device_sdk_c_build.sh
```
eb-oss, line 12: This should be added as a builder (probably to the appropriate base job), not as a post-build script.

eb-oss (changes requested):

jjb/device/device-sdk-c.yaml
ernestojeda: Sure thing

jjb/device/device-sdk-c.yaml
ernestojeda: I dont know how to do that. Do you have examples of other jobs that do this?

shell/device_sdk_c_build.sh
ernestojeda: No, you dont have to do that with docker. Docker will automatically infer environment variables for you so you dont have to do things like: `-e UID=$UID -e GID=$GID`. This has been tested on the sandbox. https://jenkins.edgexfoundry.org/sandbox/job/device-sdk-c-v2-master-verify-docker/

jjb/device/device-sdk-c.yaml
ernestojeda: Not sure as to how I would test this on the sandbox. I assume artifact publishing would occur in the "merge" job and that is not allowed on the sandbox. I saw no reference to `artifact-version` in the `jjb/edgex-templates-docker.yaml` file. Not sure if it spider's up all the way up to the LF managed templates though. Hopefully you have knowledge in that area.

jjb/device/device-sdk-c.yaml
ernestojeda: The base job in this case, `jjb/edgex-templates-docker.yaml`, is shared by a score of other jobs. This is a special case as the developer's have their own specialized build flow. I do not think it would be appropriate to add this one-off code the the base template. 🤷‍♂️

ernestojeda:
The verify jobs have passed on the sandbox for this template. You can see the results here:
https://jenkins.edgexfoundry.org/sandbox/job/device-sdk-c-v2-master-verify-docker/
https://jenkins.edgexfoundry.org/sandbox/job/device-sdk-c-v2-master-verify-docker-arm/

I had to put `-v2` in the name because another device-sdk-c job existed and we could not delete it.

shell/device_sdk_c_build.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+set -ex -o pipefail
+
+echo "--> device_sdk_c_build.sh"
+
+userId=`id -u`
```
ernestojeda, line 6: Something seems to have changed since my last build, `UID` and `GID` are no longer overridable in the script. https://jenkins.edgexfoundry.org/sandbox/job/device-sdk-c-v2-master-verify-docker/16/console. I had to rename the environment variables to get passed the error.

shell/device_sdk_c_build.sh
```diff
@@ -0,0 +1,19 @@
...
+groupId=`id -g`
+
+mkdir $WORKSPACE/release
+
+DOCKER_BUILD_CONTAINER=${DOCKER_NAME}-builder-${BUILD_ID}
+
+echo "Starting build container and generating tar.gz"
+# --privileged is required because for some reason when docker bind mounts the
+# on the workspace on the Jenkins build server, it is Read-Only
+docker run --rm -e UID=$userId -e GID=$groupId --privileged -v $WORKSPACE/release:/edgex-c-sdk/build/release ${DOCKER_IMAGE}
```
ernestojeda, line 16: See comment on line 6

jjb/device/device-sdk-c.yaml
eb-oss: Sorry, nevermind. This feature was confusingly named (post-build is a stage of the Jenkins job, but this is simply a build script that runs last during the build stage), but it does what you want it to do without violating how the Jenkins job should be formatted. This is fine.

eb-oss (approved):

PR #278, opened 2019-02-04
add the device-grove-c template

ernestojeda:

### DIFF
--- a/jjb/device/device-grove-c.yaml
+++ b/jjb/device/device-grove-c.yaml
@@ -1,0 +1,29 @@
+---
+
+- project:
+    name: device-grove-c
+    project-name: device-grove-c
+    project: device-grove-c
+    mvn-settings: device-grove-c-settings
+    docker_name: docker-device-grove-c-arm
+    docker_root: ''
+    docker_build_args: '-f scripts/Dockerfile.alpine-3.8'
+    archive-artifacts: ''
+    artifact-version: '0.1.0-SNAPSHOT'
+    build-node: centos7-docker-4c-2g
+    cron: 'H 11 * * *'
+    stream:
+      - 'master':
+          branch: 'master'
+          docker_tag: '1.0.0'
+
+    jobs:
+      - '{project-name}-{stream}-verify-docker-arm':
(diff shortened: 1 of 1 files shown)


ernestojeda:
Adding the template @lranjbar and I worked on. @MightyNerdEric @jamesrgregg  please add reviewers.

ernestojeda:
@MightyNerdEric Here is the successful verify job: https://jenkins.edgexfoundry.org/sandbox/job/device-grove-c-master-verify-docker-arm/1/console

jjb/device/device-grove-c.yaml
```diff
@@ -0,0 +1,29 @@
...
+- project:
+    name: device-grove-c
+    project-name: device-grove-c
+    project: device-grove-c
+    mvn-settings: device-grove-c-settings
+    docker_name: docker-device-grove-c-arm
+    docker_root: ''
+    docker_build_args: '-f scripts/Dockerfile.alpine-3.8'
+    archive-artifacts: ''
+    artifact-version: '0.1.0-SNAPSHOT'
```
eb-oss, line 12: Does this need to be here? I seem to recall that we pulled this line out of the last C job with no ill effects.

jjb/device/device-grove-c.yaml
```diff
@@ -0,0 +1,29 @@
...
+    build-node: centos7-docker-4c-2g
+    cron: 'H 11 * * *'
```
eb-oss, line 14: Does this need to be here? This will cause it to run daily verify and merge jobs on the master branch.

jamesrgregg: @MightyNerdEric that same chron is set for every single one of the device jobs. I checked device-random, device-mqtt etc..

SteveOss (approved):
LGTM

jjb/device/device-grove-c.yaml
lranjbar: This line is in the other docker based jobs, my opinion is we should leave it until we archive the java jobs and clean up the docker template.

jamesrgregg (approved):
Passes all Verify tests and the comments address feedback.

PR #281, opened 2019-02-08
Add snap release job

anonymouse64:
This job is meant to be manually run, providing the snap revision and channel to release the snap to. 

See edgex-go issue https://github.com/edgexfoundry/edgex-go/issues/1079 for more details and edgex-go PR https://github.com/edgexfoundry/edgex-go/pull/950 for the corresponding change to the build scripts that will enable this job.

### DIFF
--- a/jjb/edgex-go/edgex-go-snap.yaml
+++ b/jjb/edgex-go/edgex-go-snap.yaml
@@ -18,6 +18,8 @@
     jobs:
      - '{project-name}-{stream}-stage-snap-arm':
          build-node: ubuntu18.04-docker-arm64-4c-2g
+     - '{project-name}-release-snap':
+         build-node: centos7-docker-4c-2g
      - '{project-name}-{stream}-stage-snap':
          build-node: centos7-docker-4c-2g
      - '{project-name}-{stream}-verify-snap-arm':
--- a/jjb/edgex-templates-snap.yaml
+++ b/jjb/edgex-templates-snap.yaml
@@ -368,3 +368,29 @@
       - lf-provide-maven-settings-cleanup
       - edgex-provide-snap-cleanup
       - edgex-provide-docker-cleanup
+- job-template:
+    name: '{project-name}-release-snap'
+
+    # Job template for Snap release job
+    #
+    # The purpose of this job template is to run snapcraft release
+
+    <<: *snap_job_boiler_plate
+    # yamllint disable-line rule:key-duplicates
+    <<: *snap_merge_boiler_plate
+
+    parameters:
+      - string:
+          name: SNAP_CHANNEL
+          description: 'Snap channel to release to.'
+      - string:
+          name: SNAP_REVISION
+          description: 'Snap revision to release.'
+
+    builders:
+      - lf-jacoco-nojava-workaround
+      - shell: '{obj:pre_build_script}'
+      - shell: '{obj:build_script}'
+      - shell: '{obj:post_build_script}'
+      - edgex-provide-snap-cleanup
+      - edgex-provide-docker-cleanup


anonymouse64:
I can't request reviews on this PR through GitHub, but if @jamesrgregg @MightyNerdEric and @tonyespy could review this that would be great

eb-oss:
@anonymouse64 Do you have sandbox access? If so, has this been tested in the sandbox?

anonymouse64:
I do not have sandbox access and as such it hasn't been tested in the sandbox

jamesrgregg:
It would be good for Ian to get access to the sandbox. We want developers to start owning more of the Jenkins jobs.

Sent from my iPhone

On Feb 8, 2019, at 3:32 PM, Ian Johnson <notifications@github.com<mailto:notifications@github.com>> wrote:


I do not have sandbox access and as such it hasn't been tested in the sandbox

anonymouse64:
So I pushed this job up to the sandbox and ran a test version of it to release an already released snap revision, and the job seems to have failed because it can't pull the EdgeX credentials file in. To me that's fine, as on the real job (on normal jenkins.edgexfoundry.org) it should be able to pull the file in and do the release.

For reference here's the test run I did: https://jenkins.edgexfoundry.org/sandbox/job/edgex-go-snap-release-snap/1/console

jamesrgregg (commented):
The error on the sandbox is: 
16:38:46 ERROR: not able to provide the file [ManagedFile: id=EdgeX, targetLocation=$HOME/EdgeX, variable=], can't be resolved by any provider - maybe it got deleted by an administrator?
@MightyNerdEric Is this the settings file?  Why would it be deleted from the sandbox?  I'm not familiar with this error and just seeking clarification.

It would seem that we would want to persist the settings in all environments if we are going to ask developers to confirm jobs work in sandbox before PR approval.
Thank you @anonymouse64 for deploying the job to the sandbox.

anonymouse64:
I would assume that file only exists in the normal Jenkins because that
file has been successfully used in recent jobs to release to the edge
channel in daily builds. The only issue I have with putting this file into
the sandbox Jenkins is that if someone had access to the sandbox and they
printed off the contents of that file they could upload malware or
something in place of the snap as that credentials file is effectively full
permission for snap upload.

jjb/edgex-templates-snap.yaml
```diff
@@ -368,3 +368,37 @@
       - lf-provide-maven-settings-cleanup
       - edgex-provide-snap-cleanup
       - edgex-provide-docker-cleanup
+- job-template:
+    name: '{project-name}-release-snap'
+
+    # Job template for Snap merge jobs
```
tonyespy: Is this cut&paste cruft?

jjb/edgex-templates-snap.yaml
```diff
@@ -368,3 +368,37 @@
...
+    #
+    # The purpose of this job template is to run snap build
```
tonyespy: Isn't the purpose of this job template to run` snapcraft release`?

tonyespy (commented):
Two minor comments/questions...

jjb/edgex-templates-snap.yaml
anonymouse64: Yes I just copied the template I can remove this comment if you like

jjb/edgex-templates-snap.yaml
anonymouse64: copy & paste issue again I'll change this

tonyespy (dismissed):
Thanks for addressing my comments.  LGTM

jamesrgregg:
@MightyNerdEric the job was created on the sandbox but had an issue related to the settings file.  What's needed to confirm that this is ready for approval?

jjb/edgex-templates-snap.yaml
```diff
@@ -368,3 +368,33 @@
...
+      - string:
+          name: SNAP_CHANNEL
+          description: 'Snap channel to release to.'
+      - string:
+          name: SNAP_REVISION
+          description: 'Snap revision to release.'
+
+    builders:
+      - lf-jacoco-nojava-workaround
+      - lf-provide-maven-settings:
```
eb-oss: These settings files provide docker credentials. I don't see them being used anywhere, and the job seems to run fine on the sandbox without them. I would suggest removing lines 392-394 and 398.

jjb/edgex-go/edgex-go-snap.yaml
```diff
@@ -18,6 +18,8 @@
     jobs:
      - '{project-name}-{stream}-stage-snap-arm':
          build-node: ubuntu18.04-docker-arm64-4c-2g
+     - '{project-name}-release-snap':
```
eb-oss, line 21: Given that there is no stream listed here, I assume this is meant to only run on master?

eb-oss (changes requested):

eb-oss:
> @MightyNerdEric the job was created on the sandbox but had an issue related to the settings file. What's needed to confirm that this is ready for approval?

Generally, I'd ask that after the job is pushed to the sandbox, build steps that require prod settings files be manually removed (wherever possible) in order to test the job logic. For this job, I did that myself, and am running some tests right now.

jjb/edgex-go/edgex-go-snap.yaml
anonymouse64: Well technically the job doesn't run on any branch, it's meant more for _releasing_ an already built snap artifact. The snap artifacts are uploaded daily to the `edge`, `delhi/edge`, etc. channels and we manually promote individual artifacts from `edge` to `stable` or `delhi/stable`, etc. Previously this step was manual and required someone from LF to SSH into the machine and run a command using the credentials file discussed earlier.

jjb/edgex-templates-snap.yaml
anonymouse64: The snap job runs inside a docker container but doesn't release anything to dockerhub, nexus, etc. so if these credentials are for releasing things like that, then these credentials can be removed.

eb-oss: Yes, they are for Nexus repos and dockerhub.

anonymouse64: Understood, I removed the `lf-provide-maven-settings` and `lf-provide-maven-settings-cleanup` from this job template.

anonymouse64:
@MightyNerdEric this may be an issue with the job logic, but it appears as if the job you triggered ran incorrectly, as it triggered a _build_ of the snap, rather than an attempt to release an already built snap on the store. 

My logic in edgex-go's build.sh script here: https://github.com/edgexfoundry/edgex-go/blob/20ab82b8d89d1889f8feba198e45c3ec435f09a9/snap/build.sh#L22-L26 is meant to populate the env var `JOB_TYPE` to be `release` when this job, and then the docker entrypoint script here: https://github.com/edgexfoundry/edgex-go/blob/20ab82b8d89d1889f8feba198e45c3ec435f09a9/snap/entrypoint.sh#L46-L51 should not attempt to build the snap (if the credentials file is missing then it should still attempt to do the release and fail to login to the store with the credentials).

It's not clear to me right now looking at it whether the issue is with this job written in the PR, with your modifications to this job, or with my script in edgex-go. I did confirm locally that it attempts to do the local build correctly if I define `JOB_NAME=edgex-go-snap-release-snap` so I don't think the logic is with my script in edgex-go.

eb-oss:
Looks like the problem is here: https://github.com/edgexfoundry/edgex-go/blob/20ab82b8d89d1889f8feba198e45c3ec435f09a9/snap/build.sh#L24

`elif [[ "$JOB_NAME" =~ edgex-go-snap-.*-release-snap.* ]]; then`

The job name contains "snap-release-snap", which won't match that regex (it's looking for something like "snap--release-snap"). Not sure how that's working for you locally, but it looks like a bad regex to me.

anonymouse64:
ah yes you're right there was an error in that regex. I will update it in a PR to edgex-go.

anonymouse64:
@MightyNerdEric I rebased this on top of master and fixed the typo in edgex-go. This should be ready to re-review again

eb-oss (approved):

PR #282, opened 2019-02-08
Docker template cleanup after archiving Java repos

lranjbar:
This touches a lot of jobs, but doesn't make any changes to a majority of them.
I've been testing them on the sandbox. Will likely have to re-upload the changes after the sandbox wipe tonight.

Changes:
- Created custom_triggers variable in docker template
- Moved current triggers into job definitions
- Added an if statement to look for a VERSION file before using pom.xml in docker_build.sh
- Added Github trigger to device-grove-c and device-sdk-c jobs

@JPWKU @jamesrgregg @ernestojeda @MightyNerdEric 


jjb/core/core-consul.yaml
```diff
@@ -19,18 +19,30 @@
           docker_tag: 'master'
     jobs:
       - '{project-name}-{stream}-verify-docker'
-      - '{project-name}-{stream}-merge-docker'
-      - '{project-name}-{stream}-release-version-docker-daily-no-sonar'
+      - '{project-name}-{stream}-merge-docker':
+          custom_triggers:
+            - reverse:
+                jobs: '{project-name}-{stream}-merge-java'
```
eb-oss: This job does not appear to exist. How can it use a reverse trigger?

eb-oss (changes requested):
Using the custom triggers object is an interesting idea, but I'd probably prefer to create different job types that have the triggers standardized. But I'm not sure that's even necessary here; it looks like most of those reverse-triggers are referencing jobs that don't exist. But I also might be missing something that's creating those jobs (one of my biggest gripes with JJB is that the traceability is insanely low). So please let me know what you think about just creating one or two standard jobs, and what the deal is for the reverse triggers.

jjb/core/core-consul.yaml
lranjbar: Possibly more clean up needed here. If you look at production this is the current state for these jobs like core-consul, edgex-mongo, etc. Since there is no upstream job anymore (it probably existed in the past) it hasn't triggered in the past 7+ months. My goal was to minimize the changes before others could see.

The only repo where the chains of jobs all exist is support-rulesengine so at the very least support-rulesengine will need the old java triggers intact.

jamesrgregg: @MightyNerdEric the jobs that don't exist is related to the cleanup of the legacy java repos and the removal of all of the legacy jobs that were related to those java builds.  This cleanup is necessary and was under the consult of @JPWKU

jamesrgregg: consul-core is Archived

lranjbar:
I made some changes today based off feedback. Let me know if there is any more feedback to incorporate.

- Removed core-consul jobs since the repository is archived
- Changed the edgex-mongo and edgex-volume jobs to trigger off changes in the GitHub repos

These jobs are still waiting for verification on the sandbox. We will verify these once the settings files are available on the sandbox.

jjb/device/device-grove-c.yaml
```diff
@@ -15,6 +15,7 @@
     stream:
       - 'master':
           branch: 'master'
+          docker_tag: 'master'
```
JPWKU: Have 2 `docker_tag` variables here...only need one.

JPWKU (dismissed):
So far I only see an issue with the duplicate variable I noted.

lranjbar:
@MightyNerdEric This last push should take care of the feedback and what we discussed today.

lranjbar:
@MightyNerdEric @jamesrgregg 

These changes have the working verify jobs on the sandbox:
- device-grove-c-master-verify-docker-arm
- device-sdk-c-master-verify-docker
- device-sdk-c-master-verify-docker-arm
- docker-edgex-mongo-master-verify-docker
- docker-edgex-mongo-master-verify-docker-arm
- docker-edgex-volume-master-verify-docker
- docker-edgex-volume-master-verify-docker-arm
- support-rulesengine-master-verify-docker
- support-rulesengine-master-verify-docker-arm

jamesrgregg:
LGTM

jjb/volume/edgex-volume.yaml
```diff
@@ -25,17 +25,20 @@
 
     jobs:
       - '{project-name}-{stream}-verify-docker'
-      - '{project-name}-{stream}-merge-docker'
-      - '{project-name}-{stream}-release-version-docker-daily-no-sonar'
+      - '{project-name}-{stream}-merge-docker':
+          custom_triggers:
```
eb-oss: The custom triggers in this file need to be removed.

eb-oss (changes requested):

jjb/volume/edgex-volume.yaml
lranjbar: Done 👍

eb-oss (approved):

PR #286, opened 2019-02-12
Correct allure file path

eric-r-wright:
Signed-off-by: eric-r-wright <eric.r.wright@intel.com>


eb-oss (approved):
It would have been easier to just update the previous PR, but otherwise, approved.

PR #288, opened 2019-02-14
Add daily release docker jobs for mongo

JPWKU:
Also modify docker template to allow for rulesengine docker job
to still work until we rewrite in golang.

### DIFF
--- a/jjb/edgex-templates-docker.yaml
+++ b/jjb/edgex-templates-docker.yaml
@@ -240,11 +240,12 @@
       - shell: !include-raw: ../shell/docker-push.sh
 
     triggers:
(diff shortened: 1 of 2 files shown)


JPWKU:
Until we migrate rulesengine from java to go we need to keep this reverse trigger in place.  Jobs that don't have an upstream maven stage job will ignore this trigger.

JPWKU:
The daily release docker template pushes containers to the nexus staging repo 10004 daily as well in anticipation of release.

eb-oss (approved):

lranjbar:
@JPWKU Will we need this for docker-edgex-volume as well?

JPWKU:
@lranjbar Yes; docker-edgex-volume should just use the same templates and be good to go.

PR #289, opened 2019-02-14
Remove docker_tag variable from mongo and volume

JPWKU:
Only removed on master branch, this will allow the tag to be specified
in the VERSION file.

Signed-off-by: Jeremy Phelps <iot.engineering@dell.com>


xmlviking (approved):
👍 Looks good to me the docker-edgex-volume and docker-mongo-go have been PR merged with the VERSION files in them.

eb-oss (approved):

PR #313, opened 2019-02-21
Fix default GO_VERSION to 1.11

ernestojeda:

### DIFF
--- a/shell/install_custom_golang.sh
+++ b/shell/install_custom_golang.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 echo "--> install_custom_golang.sh"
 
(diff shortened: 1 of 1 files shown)


tsconn23:
@ernestojeda Are you sure this is pinned to the latest tagged release and not the initial 1.11 release?

tsconn23:
The 1.11 and 1.11.5 files are two different sizes. I don't think this change does what we want it to do.
screen shot 2019-02-21 at 3 48 15 pm

ernestojeda:
Yeah, this is what go version gives me
```
$ go version
go version go1.11 linux/amd64
```

PR #318, opened 2019-02-26
WIP: Add sigul signing to golang binaries

JPWKU:
this will sign all of the files copied into $bin_dir: 
https://lf-releng-tools.readthedocs.io/en/latest/commands/sign.html#sigul

Signed-off-by: Jeremy Phelps <jeremyphelps@linux.com>

### DIFF
--- a/shell/edgexfoundry-device-modbus-go-docker-push.sh
+++ b/shell/edgexfoundry-device-modbus-go-docker-push.sh
@@ -65,6 +65,8 @@ bin_dir=device-modbus-go-$VERSION/
 
 find cmd/ -type f -exec cp {} $bin_dir \;
 
+lftools sign sigul $bin_dir
+
 set +x
 case $DEPLOY_TYPE in
   'snapshot')
--- a/shell/edgexfoundry-device-mqtt-go-docker-push.sh
+++ b/shell/edgexfoundry-device-mqtt-go-docker-push.sh
@@ -65,6 +65,8 @@ bin_dir=device-mqtt-go-$VERSION/
 
 find cmd/ -type f -exec cp {} $bin_dir \;
 
+lftools sign sigul $bin_dir
+
 set +x
 case $DEPLOY_TYPE in
   'snapshot')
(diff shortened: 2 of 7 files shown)


JPWKU:
Note that this signs the artifacts, which is separate from signing the git tag; that functionality still needs to be built into lftools.

eb-oss:
Has this been tested on the sandbox?

JPWKU:
@MightyNerdEric No, it will fail on sandbox before it gets to the lftools call because it does a docker push first.  Not really sure of a good way to test.

jamesrgregg (dismissed):
LGTM

jamesrgregg:
> Note that this signs the artifacts, which is separate from signing the git tag; that functionality still needs to be built into lftools.

I question why we would need to sign the git tag.

eb-oss:
> I question why we would need to sign the git tag.

Signing a version tag is a way of letting users know that a release version is official and verified.

JPWKU:
Please do not merge this yet...it will break CI.

jamesrgregg:
@JPWKU is this going in for Edinburgh release ?  We got the confirmation after a lot of testing that the git repo signing is in lftools / sigul but it requires a special undocumented flag.

JPWKU:
@jamesrgregg This isn't a requirement, but would be a nice to have.  It would mean that LF would not have to manually sign the artifacts.  I don't know anything about that flag, can you catch me up on it?
Just to reiterate, the release can still be executed without this.  I also haven't tested this anywhere.

lranjbar (changes requested):
@JPWKU So I have a couple of thoughts... I know its pretty late in the release. For Edinburgh I think we only needed signed git tags because we were only targeting the go modules. IMO this should be moved to Fuji scope due to that scope focus that we decided on.

- This is to sign any artifacts in a binary directory. There is a macro in global-jjb called [lf-sigul-sign-dir](https://github.com/lfit/releng-global-jjb/blob/master/jjb/lf-macros.yaml#L284) we could use to do this too.
- In order to use the macro (Which will install and set up sigul, this is why this will fail currently. We need to set up sigul to be used) we need to set the environment variables SIGN_DIR and SIGN_MODE
- With my testing of sigul and signing directories I know that if there are no files in the directory it will fail the build. So we have to make sure there will be files in the directory to sign too.

JPWKU:
After consulting with @jamesrgregg; closing this PR so that DevOps team can re-scope in light of major changes that have happened in ci-man repo since this was opened.

PR #324, opened 2019-03-07
Add ~/.local/bin to path for edgex-publish-docs

eb-oss:
This script uses lftools, which installs into a directory that may
not be in the default PATH.


jamesrgregg (approved):
LGTM.. thank you for fixing this.

PR #333, opened 2019-03-22
Update global-jjb to v0.34.0, use infra-pre-build

eb-oss:
Using lf-infra-pre-build will install lftools, rather than installing
it manually in edgex-publish-docs.sh. Path is fixed by using "bash
-l".


eb-oss:
This fix was tested here: https://jenkins.edgexfoundry.org/sandbox/view/All/job/edgex-go-docs-delhi-stage-docs/10/

jamesrgregg (approved):
LGTM

PR #335, opened 2019-03-22
Add lf-infra-pre-build to merge-docs template

eb-oss:
This will install lftools, which is necessary for the job to
complete.


jamesrgregg (approved):
LGTM

PR #339, opened 2019-04-11
add git-semver merge build, pipeline style

dweomer:
Please note that I have disabled creation of the verify job as I am worried about the potential for malicious PRs being executed (since this job uses the `Jenkinsfile` in the repo).

Tested a verify job in the sandbox, using lots of replays: https://jenkins.edgexfoundry.org/sandbox/view/All/job/git-semver-verify-pipeline/

### DIFF
--- a/jjb/git-semver/git-semver.yaml
+++ b/jjb/git-semver/git-semver.yaml
@@ -1,0 +1,10 @@
+---
+- project:
+    name: git-semver
+    project-name: git-semver
+    project: git-semver
+    mvn-settings: git-semver-settings
+    jenkins_file: 'Jenkinsfile'
+    stream: {}
+    jobs:
+      - '{project-name}-merge-pipeline'


jjb/ci-management/git-semver.yaml
```diff
@@ -0,0 +1,12 @@
+# git-semver.xml
```
eb-oss: Wrong file extension. I don't actually see the need for this line at all.

jjb/ci-management/git-semver.yaml
```diff
@@ -0,0 +1,12 @@
...
+---
+- project:
+    name: git-semver
+    project-name: git-semver
+    project: git-semver
+    # mvn-settings: git-semver-settings
```
eb-oss: This can be uncommented. Whether you use it or not, the file is present.

jjb/ci-management/git-semver.yaml
```diff
@@ -0,0 +1,12 @@
...
+    jenkins_file: 'Jenkinsfile'
+    stream: {}
+    jobs:
+      - '{project-name}-merge-pipeline'
+#      - '{project-name}-verify-pipeline'
```
eb-oss: Go ahead and pull this out if you don't want verify jobs.

eb-oss (changes requested):
I believe this file should probably be in jjb/git-semver (which will need to be created) rather than jjb/ci-management.

jjb/ci-management/git-semver.yaml
dweomer: dunno where i picked that up from

eb-oss (dismissed):
LGTM, but I'd like one more set of eyes on this before merging, since I'm not terribly familiar with the pipeline work right now.

dweomer:
I have tested the pipeline with a temporary verify job in the sandbox: https://jenkins.edgexfoundry.org/sandbox/view/All/job/git-semver-verify-pipeline/66

ernestojeda (approved):
LGTM

lranjbar (approved):

eb-oss (approved):

PR #346, opened 2019-04-17
pipeline: revert permit-all to true

dweomer:
this should sidestep build errors that look like:

```
ERROR: No flow definition, cannot run
```


eb-oss:
Will authorized users still be limited to {github-org}?

dweomer:
> Will authorized users still be limited to {github-org}?

I believe so, but will need to verify.

jamesrgregg (approved):
LGTM

eb-oss (approved):

PR #347, opened 2019-04-17
add ci-build-images jjb template and add stream job templates back in…

ernestojeda:
Update `jjb/edgex-templates-pipeline.yaml` to add stream templates back in and remove ARM jobs as they are now part of the Pipeline script.

Add `jjb/build-images/ci-build-images.yml` to setup job templates for golang, gcc, and kong base images

### DIFF
--- a/jjb/build-images/build-images.yml
+++ b/jjb/build-images/build-images.yml
@@ -1,0 +1,18 @@
+---
+- project:
+    name: ci-build-images
+    project-name: ci-build-images
+    project: ci-build-images
+    mvn-settings: ci-build-images-settings
+    branch: master
+    stream:
+      - 'golang':
+          jenkins_file: golang/Jenkinsfile
+      - 'gcc':
+          jenkins_file: gcc/Jenkinsfile
+      - 'kong':
+          jenkins_file: kong/Jenkinsfile
+
+    jobs:
+      - '{project-name}-{stream}-verify-pipeline'
+      - '{project-name}-{stream}-merge-pipeline'
--- a/jjb/common-views.yaml
+++ b/jjb/common-views.yaml
@@ -37,3 +37,12 @@
     filter-queue: false
     recurse: false
     regex: '.*docs.*'
+
+- view:
+    name: Ci-Build-Images
+    description: 'List of Jenkins build image jobs'
+    view-type: list
+    filter-executors: false
+    filter-queue: false
+    recurse: false
+    regex: '.*-ci-build-images-.*'
--- a/jjb/edgex-templates-pipeline.yaml
+++ b/jjb/edgex-templates-pipeline.yaml
@@ -113,6 +113,17 @@
     build-node: centos7-docker-4c-2g
 
 - job-template:
+    name: '{project-name}-{stream}-verify-pipeline'
+
+    # Job template for pipeline verify jobs 
+
+    <<: *pipeline_job_boiler_plate
+    # yamllint disable-line rule:key-duplicates
+    <<: *pipeline_verify_boiler_plate
+
+    build-node: centos7-docker-4c-2g
+
+- job-template:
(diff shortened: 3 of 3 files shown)


jamesrgregg (dismissed):
LGTM - supports stream type builds

jjb/build-images/ci-build-images.yml
```diff
@@ -0,0 +1,19 @@
+---
+- project:
+    name: ci-build-images
+    project-name: ci-build-images
+    project: ci-build-images
+    mvn-settings: ci-build-images-settings
+    github-org: edgexfoundry
```
eb-oss: This is a default value (in jjb/defaults.yaml). It's not necessary to define unless you're changing it.

eb-oss (changes requested):
This needs a view added. Given that the directory name is build-images, and the file name is ci-build-images.yml, are you expecting to have other <category>-build-images? If so, it should probably be defined in a separate `build-images.yml` file (and if not, the directory name should be changed).

This commit completely removes ARM builds. Are they no longer supported? If so, that should be mentioned in the commit message.
Speaking of which, the commit message formatting should also be fixed. Please keep the subject line under 50 characters, and expand upon it in the body. LF follows the standards outlined here: https://chris.beams.io/posts/git-commit/

ernestojeda:
> This needs a view added. Given that the directory name is build-images, and the file name is ci-build-images.yml, are you expecting to have other -build-images? If so, it should probably be defined in a separate `build-images.yml` file (and if not, the directory name should be changed).

Agreed, we will have many jobs for build images. I will add a view. However, for build-images, we are trying to only have one yml file because there is really no difference between jobs other than the scriptPath. We are multiplexing the Jenkins jobs by using the `stream` feature.

> This commit completely removes ARM builds. Are they no longer supported? If so, that should be mentioned in the commit message.
> Speaking of which, the commit message formatting should also be fixed. Please keep the subject line under 50 characters, and expand upon it in the body. LF follows the standards outlined here: https://chris.beams.io/posts/git-commit/

Sure l will push a new commit message.

We fully support any CPU architecture. We have just moved the decision about architectures into the the the Jenkinsfile's themselves. We don't need dedicated ARM jobs, the CPU architecture decisions can now be handled in the Jenkinsfile by using something like this:

```Groovy
pipeline {
    //x86_64 builds here
    agent {
        label 'centos7-docker-4c-2g'
        ... rest of the build code
    }

    //arm_64 builds here
    agent {
        label 'ubuntu18.04-docker-arm64-4c-2g'
        ... rest of the build code
    }
}
```

You can see examples of that see the [git-semver](https://github.com/edgexfoundry/git-semver/blob/master/Jenkinsfile#L33-L49) repo or the [sample-service](https://github.com/edgexfoundry-holding/sample-service/blob/master/Jenkinsfile#L38-L64) repo.

jjb/build-images/ci-build-images.yml
ernestojeda: All of these are defaults...or just `github-org`

eb-oss: Just `github-org`.

eb-oss:
> Agreed, we will have many jobs for build images. I will add a view. However, for build-images, we are trying to only have one yml file because there is really no difference between jobs other than the scriptPath. We are multiplexing the Jenkins jobs by using the stream feature.

I'm not picky about the name, I just want to make sure that there's consistency. If there's only going to be one file named `ci-build-images`, it seems to me that the directory should also be titled `ci-build-images`. Or the other way around, where if you'll define more than just CI build images in that one file, the file should be named `build-images.yml`.

ernestojeda:
@MightyNerdEric Let me know if there is anything else that needs to be done before this is merged.

eb-oss (approved):

PR #349, opened 2019-04-19
allow whitelisted orgs to be ghprb 'admins'

dweomer:
this should allow edgexfoundry members to 'recheck' PRs submitted by
non-members.

also, tweaked the git-semver jobs to have a view for easy finding upon
logging into jenkins.

### DIFF
--- a/jjb/edgex-templates-pipeline.yaml
+++ b/jjb/edgex-templates-pipeline.yaml
@@ -48,9 +48,6 @@
 
 - pipeline_verify_boiler_plate: &pipeline_verify_boiler_plate
     name: pipeline_verify_boiler_plate
-
-    concurrent: true
-
     pipeline-scm:
       script-path: '{jenkins_file}'
       scm:
@@ -67,11 +64,12 @@
       - github-pull-request:
           trigger-phrase: '^recheck$'
           only-trigger-phrase: false
-          status-context: '{status-context}'
+          status-context: '{project-name}-verify-pipeline'
           permit-all: false
           github-hooks: true
           org-list:
             - '{github-org}'
+          allow-whitelist-orgs-as-admins: true
           auto-close-on-fail: false
           white-list-target-branches:
             - '{branch}'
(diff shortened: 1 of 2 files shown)


jjb/edgex-templates-pipeline.yaml
```diff
@@ -67,19 +65,20 @@
...
+          allow-whitelist-orgs-as-admins: true
           auto-close-on-fail: false
           white-list-target-branches:
             - '{branch}'
 
 
 - pipeline_merge_boiler_plate: &pipeline_merge_boiler_plate
     name: pipeline_merge_boiler_plate
-
+    # concurrent: true
```
dweomer: this was helpful in side-by-side comparison in split view.

also, @lranjbar, should we consider turning this on?

jjb/git-semver/git-semver.yaml
```diff
@@ -10,3 +10,5 @@
     jobs:
       - '{project-name}-merge-pipeline'
       - '{project-name}-verify-pipeline'
+    views:
+      - project-view
```
dweomer, line 14: based on the other templates, this looked like it should work.

eb-oss (dismissed):
LGTM, waiting on feedback from @lranjbar.

jjb/edgex-templates-pipeline.yaml
lranjbar: Should be fine to turn this back on. When I turned this off the template still was building from two Jenkinsfiles for verify and merge and this was to guard against a potential race condition.

dweomer: in my update, i nixed this line from the merge boilerplate as well as from the verify boilerplate because it is set to true in the common job boilerplate

jamesrgregg (dismissed):
LGTM

dweomer:
With the manual change made to the git-semver repo this morning and this issue that @lranjbar found (https://support.cloudbees.com/hc/en-us/articles/115003015691-GitHub-Webhook-Non-Multibranch-Jobs) we should be done with the merge job trigger issue and these should be some of the last changes required to get the git-semver pipeline to cut and push release images.

jamesrgregg (approved):
LGTM

PR #352, opened 2019-04-24
allow for custom status-context to be passed

ernestojeda:
I am reverting the status-context change to allow custom status-context to be passed in at the project level. The `ci-build-images` project has 3/4 different "streams" which break when the status-context is set to `{project-name}-verify-pipeline` at the template level.

This also keeps it consistent with the other JJB templates.

### DIFF
--- a/jjb/build-images/build-images.yml
+++ b/jjb/build-images/build-images.yml
@@ -5,6 +5,7 @@
     project: ci-build-images
     mvn-settings: ci-build-images-settings
     branch: master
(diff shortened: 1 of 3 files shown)


jamesrgregg (approved):
LGTM

eb-oss:
Has this been tested to ensure you're getting the result you want?

ernestojeda:
> Has this been tested to ensure you're getting the result you want?

This is not testable on the sandbox due to webhooks not working over there. I am just reverting it to it's previous value.

PR #353, opened 2019-04-25
Add lftools image build

lranjbar:

### DIFF
--- a/jjb/build-images/build-images.yml
+++ b/jjb/build-images/build-images.yml
@@ -13,6 +13,8 @@
           jenkins_file: gcc/Jenkinsfile
       - 'kong':
           jenkins_file: kong/Jenkinsfile
+      - 'lftools':
+          jenkins_file: lftools/Jenkinsfile
 
     jobs:
       - '{project-name}-{stream}-verify-pipeline'


ernestojeda (dismissed):
LGTM

eb-oss:
Could you link to the related Jenkinsfile?

lranjbar:
It's on this branch currently:
https://github.com/lranjbar/ci-build-images/blob/lftools/lftools/Jenkinsfile

@MightyNerdEric

ernestojeda (approved):
LGTM

eb-oss:
I'm not sure I understand the necessity of this. We have macros in global-jjb to install lftools and perform sigul signing. Why create a docker container for it, when our build VMs can quickly install it directly?

lranjbar:
TLDR: I made this because lftools and sigul weren't playing nice in my test pipeline jobs.

The first reason is that the global macros are composed of freestyle job components and aren't compatible with pipeline jobs. In the the global library you can see I started to recreate some of the global macros to deal with this. [Ex: edgeXInfraPublish](https://github.com/edgexfoundry/edgex-global-pipelines/blob/master/vars/edgeXInfraPublish.groovy)

I made this image when I was testing lftools and sigul for the pipeline jobs.  With lftools and sigul initally I tried a similar strategy to install them inside the pipeline jobs just using the shell scripts and no containers. For some reason it wasn't installing them so they could call each other it was like it was installing them in separate contexts. Lots of command not found errors. When I made the container it was to force them into the same context so lftools could call sigul.  It was only after making the container was I able to call both lftools and sigul in following commands.

Beyond that I also think it it will speed up the builds eventually. Installing lftools takes about a minute to do and some of the freestyle jobs do this 2-3 times a build. This is a very small step towards optimizing builds. :)

PR #354, opened 2019-04-25
git-semver: remove local-branch from merge job

dweomer:
it wasn't working as expected, aka at all, and so setting an environment variable for git-semver to use as an override


ernestojeda (approved):
LGTM

PR #357, opened 2019-04-30
Integrate snap build script into jenkins job definition

anonymouse64:
This PR generalizes the snap build scripts we have from edgex-go(see [here](https://github.com/edgexfoundry/edgex-go/blob/master/snap/build.sh), [here](https://github.com/edgexfoundry/edgex-go/blob/master/snap/entrypoint.sh), and [here](https://github.com/edgexfoundry/edgex-go/blob/master/snap/Dockerfile.build)) and stores it all in the jenkins job definition directly as the `build_script` so that when we add more repositories, it's easy to create an associated snap job as well by just templating out the job definitions for that project/repo and specifying the `{project-name}-{stream}-stage-snap` and `{project-name}-{stream}-verify-snap`, etc. 

I deployed this job on the sandbox and was able to build the associated edgex-go PR https://github.com/edgexfoundry/edgex-go/pull/1297 successfully. See https://jenkins.edgexfoundry.org/sandbox/view/All/job/edgex-go-snap-master-verify-snap/1/console. I did modify the job definition slightly by removing the macro which adds the snapcraft credentials to the job when running because those credentials aren't available to the sandbox jenkins and they're not really needed to test the job at all. The patch is here if anyone wants to mess around with the sandbox job definition:
```patch
diff --git a/jjb/edgex-templates-snap.yaml b/jjb/edgex-templates-snap.yaml
index 20299d7..b6cee40 100644
--- a/jjb/edgex-templates-snap.yaml
+++ b/jjb/edgex-templates-snap.yaml
@@ -45,7 +45,7 @@
           files:
             - file-id: netrc
               target: '$HOME/.netrc'
-      - edgex-snap-wrapper
+      # - edgex-snap-wrapper
 
     publishers:
       - lf-infra-publish
```

The `edgexfoundry-snapcraft.sh` script we have here is not ideal, as it uses 2 heredocs to basically store 2 other scripts inside the single script so that all of that is available from the job definition directly, but here's the reasoning for why I did that:
* The Dockerfile is okay to store in the script and feed via stdin to the `docker build` process because docker build supports stdin input of the Dockerfile. This doesn't pose any complications or differences in how the docker container is built or runs. We currently need to include all of the files from the project into the docker container as well as the credentials file so that when the build is finished, the snap artifacts don't leave the container and just get pushed into the store when successful. We could use volumes but I was told that this was complicated to setup such that SELinux allowed the mounts to happen and that it was better to just inject all the files we need during the build into the container image. 
* The entrypoint.sh script is a bit more difficult because we need that script to be available to docker as part of it's build context somehow when the container is being built. We can't just store the entrypoint.sh somewhere in /tmp, etc. as it needs to be accessible inside the container. We could have the Dockerfile download the entrypoint.sh script from somewhere (perhaps even this repository), but that leads to complicated syncing problems especially when modifying that file, as you need to update the file in a way that the build when it runs to test actually is able to download that file. As such, the temporary best option is to just store that file inside the build context directly by saving the file into `$GIT_ROOT/snap/entrypoint.sh`.

I also ran into a bit of difficulty with the `edgexfoundry-snapcraft.sh` script and the `entrypoint.sh` heredoc, where the original entrypoint.sh had functions defined for various things and just called those bash functions, but I couldn't figure out how to escape the necessary `{` and `}` characters in such a way that the function definitions were still valid bash and allowed as valid job definition YAML. If I used the `include-raw` macro, it would fail to generate the job definition because it would treat the function definition body as a key to replace with a jenkins job builder variable, so I had to use `{{` and `}}` in the file to generate the appropriate job definition, but then the `{{` weren't re-replaced to be just `{` again and so it wasn't a valid bash script... I also tried numerous ways of escaping the `{` with the `include-raw-escape` macro and everytime the `{` would get replaced into `{{`, so in the end I just removed the functions and called the code directly wherever I needed to use the function. It's not great but it works.

I also setup the delhi override successfully such that we won't need to backport these changes to delhi and delhi jobs can continue to build using the build.sh script present in that repo. You can see that job on the sandbox here: https://jenkins.edgexfoundry.org/sandbox/view/All/job/edgex-go-snap-delhi-stage-snap/1/console (but not the job itself is marked as failure only because it couldn't actually release the snap)

Note that I think we should merge this into ci-management before merging the edgex-go PR so that the edgex-go PR can actually use these job definitions to test those changes.

Finally, I also fixed some typos here and removed the redundant "snap" in the arm verify job name.

### DIFF
--- a/shell/edgexfoundry-snapcraft.sh
+++ b/shell/edgexfoundry-snapcraft.sh
@@ -1,16 +1,16 @@
 #!/bin/bash -x
 
-# note that since this script gets included into a yaml string using jjb's
+# Note that since this script gets included into a yaml string using jjb's
 # include-raw-escape macro, we can't use any squiggly braces here, otherwise
-# they will get duplicated
-# for some reason we also can't use include-raw macro because that macro will
-# attempt to interpret anything contianed in a macro as a jenkins key to
-# expand which we also don't want
+# they will get duplicated in the resultant job definition.
+# For some reason we also can't use include-raw macro because that macro will
+# attempt to interpret anything contained in a macro as a jenkins key to
+# expand which we also don't want.
 
-# if we are running inside a jenkins instance then copy the login file 
-# and check if this is a release job
+# If we are running inside a jenkins instance then copy the login file and 
+# figure out what kind of job type we are running as
 if [ -n "$JENKINS_URL" ]; then
-    # since we are running from jenkins, use the workspace as the git root
+    # Since we are running from jenkins, use the workspace as the git root
     GIT_ROOT=$WORKSPACE
     if [ -f "$HOME/EdgeX" ]; then
        cp "$HOME/EdgeX" "$GIT_ROOT/edgex-snap-store-login"
@@ -18,7 +18,7 @@ if [ -n "$JENKINS_URL" ]; then
         echo "I seem to be running on Jenkins, but there's not a snap store login file..." 
     fi
 
-    # figure out what kind of job this is using $JOB_NAME and simplify that 
+    # Figure out what kind of job this is using $JOB_NAME and simplify that 
     # into $JOB_TYPE
     if [[ "$JOB_NAME" =~ .*-snap-.*-stage-snap.* ]]; then
         JOB_TYPE="stage"
@@ -28,72 +28,73 @@ if [ -n "$JENKINS_URL" ]; then
         JOB_TYPE="build"
     fi
 else
-    # not running in a jenkins container, assume the current working directory
-    # is the repo git dir that we want
+    # We're not running in a jenkins container so assume the current working 
+    # directory is the repo git dir that we want
+    # TODO: maybe support providing the dir to use as the git root as an arg
+    # or option to this script?
     GIT_ROOT=$PWD
 fi
 
-# find the name of this snap from the snapcraft.yaml "name" key
-# note that we assume that snapcraf.yaml is located in snap/snapcraft.yaml 
+# Find the name of this snap from the snapcraft.yaml "name" key.
+# Note that we assume that snapcraft.yaml is located in snap/snapcraft.yaml 
 # but it is technically allowed to store it in other locations, we just don't
 # support that here
 # TODO: implement better handling of snapcraft.yaml file discovery
 SNAP_NAME=$(grep -Po '^name: \K(.*)' snap/snapcraft.yaml)
 
-# save the entrypoint.sh script inside the build context we will send to docker
-# we have to do this because the dockerfile needs access to that file from it's
+# Save the entrypoint.sh script inside the build context we will send to docker.
+# We have to do this because the dockerfile needs access to that file from its
 # build context, so either we download the entrypoint script, or we include it
-# into the filesystem build context. downloading isn't great because then 
+# into the filesystem build context. Downloading isn't great because then 
 # there's additional linkages between this script and wherever we download from
-# and it gets complicated to manage that
-# we save it at snap/entrypoint.sh
+# and it gets complicated to manage that.
+# Note the file is saved at snap/entrypoint.sh
 (
 cat <<'EOF'
 #!/bin/bash -e
 
-# Required by click.
+# Required by the python click framework.
 export LC_ALL=C.UTF-8
 export SNAPCRAFT_SETUP_CORE=1
 
-# this tells snapcraft to include a manifest file in the snap
-# detailing which packages were used to build the snap
+# This tells snapcraft to include a manifest file in the snap
+# detailing which packages were used to build the snap.
 export SNAPCRAFT_BUILD_INFO=1
 
-# if snapcraft ever encounters any bugs, we should force it to 
+# If snapcraft ever encounters any bugs, we should force it to 
 # auto-report silently rather than attempt to ask for permission
-# to send a report
+# to send a report.
 export SNAPCRAFT_ENABLE_SILENT_REPORT=1
 
 case "$JOB_TYPE" in 
     "stage")
-        # stage jobs build the snap locally and release it
+        # Stage jobs build the snap locally and release it
         pushd /build > /dev/null
         snapcraft clean
         snapcraft
         popd > /dev/null
         pushd /build > /dev/null
         snapcraft login --with /build/edgex-snap-store-login
-        # push the snap up to the store and get the revision of the snap
+        # Push the snap up to the store and get the revision of the snap
         REVISION=$(snapcraft push "$SNAP_NAME"*.snap | grep -Po 'Revision \K[0-9]+')
-        # now release it on the provided revision and snap channel
+        # Now release it on the provided revision and snap channel
         snapcraft release "$SNAP_NAME" "$REVISION" "$SNAP_CHANNEL" 
-        # also update the meta-data automatically
+        # Also force an update of the meta-data
         snapcraft push-metadata "$SNAP_NAME"*.snap --force
         popd > /dev/null
     ;;
     "release")
-        # release jobs will promote an already built snap revision
-        # in the store to a channel
+        # Release jobs will promote an already built snap revision
+        # in the store to a channel.
         snapcraft login --with /build/edgex-snap-store-login
         snapcraft release "$SNAP_NAME" "$SNAP_REVISION" "$SNAP_CHANNEL"
     ;;
     *)
-        # do normal build and nothing else
+        # Do normal build and nothing else to verify the snap builds
         pushd /build > /dev/null
         snapcraft clean
         snapcraft
         popd > /dev/null
-
     ;;
 esac
 
@@ -102,8 +103,8 @@ EOF
 
 chmod +x "$GIT_ROOT/snap/entrypoint.sh"
 
-# build the container image - providing the relevant architecture we're on
-# to determine which snap arch to download in the docker container
+# Build the container image, providing the relevant architecture we're on
+# to determine which snap arch to download in the docker container.
 case $(arch) in 
     x86_64)
         arch="amd64";;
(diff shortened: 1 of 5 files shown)


shell/edgexfoundry-snapcraft.sh
```diff
@@ -0,0 +1,175 @@
+#!/bin/bash -x
+
+# note that since this script gets included into a yaml string using jjb's
+# include-raw-escape macro, we can't use any squiggly braces here, otherwise
+# they will get duplicated
+# for some reason we also can't use include-raw macro because that macro will
+# attempt to interpret anything contianed in a macro as a jenkins key to
```
eb-oss: *contained

shell/edgexfoundry-snapcraft.sh
```diff
@@ -0,0 +1,175 @@
...
+    GIT_ROOT=$WORKSPACE
+    if [ -f "$HOME/EdgeX" ]; then
+       cp "$HOME/EdgeX" "$GIT_ROOT/edgex-snap-store-login"
+    else
+        echo "I seem to be running on Jenkins, but there's not a snap store login file..." 
+    fi
+
+    # figure out what kind of job this is using $JOB_NAME and simplify that 
+    # into $JOB_TYPE
+    if [[ "$JOB_NAME" =~ .*-snap-.*-stage-snap.* ]]; then
```
eb-oss, line 23: Does this and the regex two lines below need to include "snap"? If the script is running, can we not assume we are in the correct job type?

shell/edgexfoundry-snapcraft.sh
```diff
@@ -0,0 +1,175 @@
...
+
+# if we are running inside a jenkins instance then copy the login file 
+# and check if this is a release job
+if [ -n "$JENKINS_URL" ]; then
+    # since we are running from jenkins, use the workspace as the git root
+    GIT_ROOT=$WORKSPACE
+    if [ -f "$HOME/EdgeX" ]; then
+       cp "$HOME/EdgeX" "$GIT_ROOT/edgex-snap-store-login"
+    else
+        echo "I seem to be running on Jenkins, but there's not a snap store login file..."
```
eb-oss, line 18: Just want to confirm that we don't want to abort here, but just echo this warning and continue on.

shell/edgexfoundry-snapcraft.sh
```diff
@@ -0,0 +1,175 @@
...
+        JOB_TYPE="build"
+    fi
+else
+    # not running in a jenkins container, assume the current working directory
+    # is the repo git dir that we want
+    GIT_ROOT=$PWD
+fi
+
+# find the name of this snap from the snapcraft.yaml "name" key
+# note that we assume that snapcraf.yaml is located in snap/snapcraft.yaml
```
eb-oss: *snapcraft.yaml
Also, this is a minor quibble, but the combination of no caps at the beginning of sentences and no periods at the end of some sentences (where EOL seems to be an implicit end of sentence) makes this a bit confusing to read. Readability could certainly be improved.

shell/edgexfoundry-snapcraft.sh
```diff
@@ -0,0 +1,175 @@
...
+fi
+
+# find the name of this snap from the snapcraft.yaml "name" key
+# note that we assume that snapcraf.yaml is located in snap/snapcraft.yaml 
+# but it is technically allowed to store it in other locations, we just don't
+# support that here
+# TODO: implement better handling of snapcraft.yaml file discovery
+SNAP_NAME=$(grep -Po '^name: \K(.*)' snap/snapcraft.yaml)
+
+# save the entrypoint.sh script inside the build context we will send to docker
```
eb-oss: Another sentence that seems to have an implicit end at EOL.

shell/edgexfoundry-snapcraft.sh
```diff
@@ -0,0 +1,175 @@
...
+# we have to do this because the dockerfile needs access to that file from it's
```
eb-oss: *its

eb-oss (changes requested):

shell/edgexfoundry-snapcraft.sh
anonymouse64: The reason not to exit is mainly just for convenience in testing on the sandbox and locally so that I can test locally by running (from the `edgex-go` dir cloned locally):
```bash
JENKINS_URL=something JOB_NAME=edgex-go-snap-master-stage-snap /home/me/git/ci-management/shell/edgexfoundry-snapcraft.sh
```
and have the job continue to run.

We could certainly make it abort here too but then to test on the sandbox and locally I would need to either create a fake file at `$HOME/EdgeX` or modify the script

shell/edgexfoundry-snapcraft.sh
anonymouse64: Yes the regex could leave out snap and still be fine, I assume you mean something like

```bash
    if [[ "$JOB_NAME" =~ .*-stage-snap.* ]]; then
        JOB_TYPE="stage"
    elif [[ "$JOB_NAME" =~ .*-release-snap ]]; then
        JOB_TYPE="release"
    else
        JOB_TYPE="build"
    fi
```

?

eb-oss: I meant if could be just `.*-stage-.*` and `.*-release-.*`

shell/edgexfoundry-snapcraft.sh
anonymouse64: Fixed (and changed all of the other comments in this file)

shell/edgexfoundry-snapcraft.sh
anonymouse64: Fixed

shell/edgexfoundry-snapcraft.sh
anonymouse64: Fixed

shell/edgexfoundry-snapcraft.sh
anonymouse64: Fixed

shell/edgexfoundry-snapcraft.sh
anonymouse64: Ah, yes it could be. Do you want me to change it to use the simpler regex?

eb-oss: To me, simpler means the code is more generalized and adaptable. My preference would be to make it as generic as possible, so long as it won't create false positives.

anonymouse64: Hmm, well I guess after thinking about it some more I would rather leave it at `*-snap-.*-stage-snap.*` and `.*-release-snap` because while yes the script won't ever do anything other than run snap jobs, we may at some point have a repo name like `device-stage` where the job names would then look like `device-stage-snap-release-snap` and `device-stage-snap-master-stage-snap` in which case the simpler regex would cause false positives. I agree with the sentiment of having simpler and generic as possible, but I think stricter is better in this instance.

eb-oss: Ok, that's acceptable to me.

eb-oss (dismissed):
I'd like to see another approval from a dev who is more familiar with the snap building process than I am before merging.

anonymouse64:
@tonyespy can review this when he gets back from Seoul

jamesrgregg (commented):
@anonymouse64 I'm not seeing any code in the `ci-snaps-common` repo.  Is there still a plan or a need to make use of that repo?

jamesrgregg:
@tsconn23 please take a look as well since we discussed this today in our Seoul F2F

anonymouse64:
@jamesgregg that repository, if needed, will be referenced from the snapcraft.yaml files in individual repositories, not from here directly. It's also not clear to me if I will need that repository at all as I didn't end up needing it for edgex-go or some of the other snaps I've prototyped with.

anonymouse64:
I modified the Dockerfile slightly to change how we "install" snaps there to be less repetitive and also to add the core18 snap there too, which while not strictly necessary is useful for snapcraft in determining missing library dependencies. I re-pushed the job to the sandbox here: https://jenkins.edgexfoundry.org/sandbox/view/All/job/edgex-go-snap-master-verify-snap/6/console and the build seems to be working (it hasn't yet finished however).

eb-oss (dismissed):

shell/edgexfoundry-snapcraft.sh
tonyespy: Did you fix by adding the apostrophe? The current usage is incorrect.

shell/edgexfoundry-snapcraft.sh
tonyespy: snapcraf.yaml --> snapcraft.yaml

tonyespy (commented):
One general comment/nit I have is that your commit message summaries all include a path to the modified file as a prefix which causes the line length to be >50 chars. Out of 300+ commits in this repo, only five commits use this kind of prefix...

Otherwise the approach and implementation in this PR seems correct to me.

Please see my two other inline comments.

And finally, looks like a pesky "merge master into <feature branch>" was introduced recently...

eb-oss:
I agree with Tony that the first line of the commit should be limited to 50 chars. You can expand on that in the description, but I also tend to think that putting the path in the commit message is unnecessary, as the commit itself holds file change info.

That "merge master into..." was due to me using the "update branch" button on this PR. I was thinking it was a rebase (as it is in Gerrit, which LF uses much more extensively than Github), so I apologize for that. Not the end of the world to have in the history, but I agree that it's annoying.

anonymouse64:
I adjusted the commit messages. FWIW, I think that looking back on history seeing the file in the summary is very useful because many times the commit makes a specific change to a specific file and referring to the change alone isn't specific enough - a similar change could have (or in some cases already has) been made in a different file from the one that the commit is making and you want to be able to quickly tell what a commit changed. Additionally, the file that was changed allows you to quickly skim down a list of commit summaries and see what files they changed - in projects that follow this pattern I can go back to the project after a lot of work has been done and just by looking at the commit summaries with their filenames see which commits might be of interest to me that I should look at. It's not a replacement for `git blame`, but a good way to skim over git history.

anonymouse64:
I also think the whole 50 chars vs 72 chars is arbitrary...

tonyespy (dismissed):
LGTM

eb-oss (dismissed):
LGTM. @jamesrgregg should I wait for @tsconn23 review before merging?

tsconn23:
@MightyNerdEric Don't wait for me given the approvals above.

eb-oss (approved):

PR #359, opened 2019-04-30
Add github webhook workaround for pipeline jobs

lranjbar:
After researching the Jenkins Github plugin and learning the limitations of it. Basically, it is unable to use the SCM block in a pipeline job to register the Github webhook. One of the workarounds I had was just to create a freestyle job that has no operation other than to register the webhook.

I've put this job up on the sandbox and it tries to register the webhook but is unable to. I suspect this is because the settings files on the sandbox don't have the correct permissions for Github.

The configuration for this job is here:
https://jenkins.edgexfoundry.org/sandbox/view/All/job/ci-build-images-pipeline-webhooks/configure

Let me know what y'all think.

@MightyNerdEric @dweomer @ernestojeda 

### DIFF
--- a/jjb/build-images/build-images.yml
+++ b/jjb/build-images/build-images.yml
@@ -19,3 +19,5 @@
     jobs:
       - '{project-name}-{stream}-verify-pipeline'
       - '{project-name}-{stream}-merge-pipeline'
+      - '{project-name}-pipeline-webhooks':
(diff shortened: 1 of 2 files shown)


ernestojeda (approved):
I don't know all the intricacies of the changes in `jjb/edgex-templates-pipeline.yaml`, but this looks good to me.

jamesrgregg (approved):
LGTM

PR #360, opened 2019-05-06
remove docker job after moving to ci-build-images

ernestojeda:
After migrating the Go and GCC base docker images over to: https://github.com/edgexfoundry/ci-build-images, the corresponding builds here in ci-management are redundant. This PR removes the no longer used docker builds.


ernestojeda:
The Docker verify checks will fail since we are removing the files. Maybe I will have to do this in two separate PR's. The first PR would only remove the JJB templates. The second PR would remove the Dockerfiles.

lranjbar (dismissed):
LGTM. 

I looked at the errors in the logs. It's failing because it can't build the images after the removal of the Dockerfiles. Should be fine because this also removes the failing jobs.

eb-oss (changes requested):
Please change commit message to imperative form.

eb-oss (approved):

PR #362, opened 2019-05-07
Change go-mod-messaging jobs to pipeline jobs

lranjbar:

### DIFF
--- a/jjb/go-modules/go-mod-messaging.yaml
+++ b/jjb/go-modules/go-mod-messaging.yaml
@@ -5,15 +5,12 @@
     project-name: go-mod-messaging
     project: go-mod-messaging
     mvn-settings: go-mod-messaging-settings
-    build_script: 'make test'
-    stream:
-      - 'master':
-          branch: 'master'
-          pre_build_script: !include-raw-escape: shell/install_custom_golang.sh
-          go-root: '/opt/go-custom/go'
+    jenkins_file: 'Jenkinsfile'
+    status-context: '{project-name}-verify-pipeline'
+    stream: {}
 
     jobs:
-    - '{project-name}-{stream}-verify-go':
-          status-context: '{project-name}-{stream}-verify'
-    - '{project-name}-{stream}-verify-go-arm':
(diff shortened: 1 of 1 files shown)


jjb/go-modules/go-mod-messaging.yaml
```diff
@@ -5,15 +5,12 @@
...
     mvn-settings: go-mod-messaging-settings
-    build_script: 'make test'
-    stream:
-      - 'master':
-          branch: 'master'
-          pre_build_script: !include-raw-escape: shell/install_custom_golang.sh
-          go-root: '/opt/go-custom/go'
+    jenkins_file: 'Jenkinsfile'
+    status-context: '{project-name}-verify-pipeline'
+    stream: {}
```
eb-oss, line 10: Most of this looks fine to me, but I'm curious about the removal of the stream. This removes the definition of `branch`, which in `build-images.yaml` is defined at the project level (see [this line](https://github.com/edgexfoundry/ci-management/blob/0c15ff97b17183ca06e5919ee5bb359e9e0ca7a4/jjb/build-images/build-images.yml#L7)). Should branch be defined in this job? And should we at least define an otherwise-empty stream?

eb-oss (commented):
Somewhat related to my inline comment about defining branch: has this been tested on the sandbox? It looks pretty close to `build-images.yaml`, but the lack of definitions for `branch` and `stream` seem like potential issues. I just want to make sure it's been run through its paces.

jjb/go-modules/go-mod-messaging.yaml
lranjbar: The overall thought is that we don't need separate jobs from different branches because the Jenkinsfile will define the different branch functionality in pipeline jobs. Collapsing all the verify and merge jobs into one job each. In most JJB templates stream is used for branches. In build-images we used it for the different images instead of branches. We have to at least define an empty object for stream or JJB will not parse it correctly and error out.

eb-oss (approved):

PR #372, opened 2019-05-16
Update centos7-builder image, remove basebuild

eb-oss:
centos7-builder and centos7-basebuild were both very out of date.
Builder is meant to replace basebuild, and basebuild is not in use
in any job, so it has been entirely removed from the repo.


eb-oss:
JJB Verify failure is expected here, since that is the issue this commit is meant to fix.

anonymouse64 (approved):
I'm not too familiar with all these changes to review them properly, but nothing looks obviously wrong here.

jamesrgregg (approved):
based on comment by @MightyNerdEric Verify is expected to fail.. which it did.

PR #378, opened 2019-05-28
JJB definitions for edgex-go edinburgh

### LINKED ISSUE #377: Add edinburgh edgex-go jobs & Remove caliornia
Modify JJB script for edgex-go job definitions to include `edinburgh` work stream and remove `california`. The latter is no longer supported.

tsconn23:
Fix #377

Adding JJB job definitions for edinburgh branch
* edgex-go
* edgex-go-docs
* edgex-go-snap

California job definitions have been removed.

Signed-off-by: Trevor Conn <trevor_conn@dell.com>

### DIFF
--- a/jjb/edgex-go/edgex-go-snap.yaml
+++ b/jjb/edgex-go/edgex-go-snap.yaml
@@ -12,6 +12,9 @@
           branch: 'delhi'
           build_script: 'cd snap && ./build.sh'
           snap-channel: delhi/edge
+      - 'edinburgh':
+          branch: 'edinburgh'
+          snap-channel: edinburgh/edge
 
     jobs:
      - '{project-name}-{stream}-stage-snap-arm':
--- a/jjb/edgex-go/edgex-go-docs.yaml
+++ b/jjb/edgex-go/edgex-go-docs.yaml
@@ -7,10 +7,10 @@
     stream:
       - 'master':
           branch: 'master'
-      - 'california':
-          branch: 'california'
       - 'delhi':
           branch: 'delhi'
+      - 'edinburgh':
+          branch: 'edinburgh'
 
     jobs:
       - '{project-name}-{stream}-verify-docs':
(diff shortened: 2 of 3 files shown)


jamesrgregg (dismissed):
LGTM..thanks @tsconn23

jjb/edgex-go/edgex-go-snap.yaml
```diff
@@ -12,6 +12,10 @@
           branch: 'delhi'
           build_script: 'cd snap && ./build.sh'
           snap-channel: delhi/edge
+      - 'edinburgh':
+          branch: 'edinburgh'
+          build_script: 'cd snap && ./build.sh'
```
anonymouse64: This is wrong, we only override `build_script` for delhi, it can be left as the default for edinburgh (i.e. don't specify `build_script` here).

anonymouse64 (changes requested):
Needs a slight change for the snap job, and a minor nitpick is that I would like to see removal of the california jobs and addition of the edinburgh jobs as two separate commits, but it's not a blocker for me.

jjb/edgex-go/edgex-go-snap.yaml
tsconn23: @anonymouse64 `build_script` line removed.

anonymouse64 (dismissed):
Snap jobs LGTM

BTW did anyone try these jobs out on the sandbox? I know that we are trying to get more devs to test things on the sandbox, and the snap job at least definitely would have failed immediately if it had been tested on the sandbox.

lranjbar (dismissed):
LGTM. Just need the edinburgh branch in edgex-go before merge in ci-management

Regarding snap jobs on the sandbox... there might be some knowledge sharing we need to do. I was able to pop the master jobs on the sandbox but not sure how to test it there. There is an extra settings file for snap jobs that might need to be mocked out. We use sandbox-settings for the "project settings" files that are named in production.

anonymouse64:
> There is an extra settings file for snap jobs that might need to be mocked out

I always just remove the reference to the settings file (though using the sandbox-settings seems fine instead though) when uploading to the sandbox and do a build of the stage-snap job with default parameters. I think that's sufficient of a test to see if the test passes/fails, but I suppose in this instance you might not even get that far since the branches don't exist yet. I assumed that the edinburgh branches had already been created, but it seems they haven't yet, so testing in the sandbox would be difficult if not impossible.

eb-oss (dismissed):

jamesrgregg (approved):
LGTM

PR #392, opened 2019-05-30
Update edgex-ui-go JJB for Edinburgh release

jamesrgregg:
Signed-off-by: James Gregg <james.r.gregg@intel.com>

### DIFF
--- a/jjb/ui/edgex-ui-go.yaml
+++ b/jjb/ui/edgex-ui-go.yaml
@@ -15,7 +15,10 @@
       - 'delhi':
           branch: 'delhi'
       - 'edinburgh': 
-          branch: 'edinburgh'    
+          branch: 'edinburgh'
+          pre_build_script: !include-raw-escape: shell/install_custom_golang.sh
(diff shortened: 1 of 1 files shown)


jjb/ui/edgex-ui-go.yaml
```diff
@@ -14,6 +14,8 @@
           go-root: '/opt/go-custom/go'
       - 'delhi':
           branch: 'delhi'
+      - 'edinburgh': 
+          branch: 'edinburgh'
```
eb-oss: It looks to me like this should have a pre_build_script, build_script, and go-root defined.

eb-oss (changes requested):

jjb/ui/edgex-ui-go.yaml
jamesrgregg: Thanks @MightyNerdEric 
I have updated as requested.

eb-oss (approved):

PR #396, opened 2019-05-31
edgexfoundry-snapcraft.sh: use --release option to push

anonymouse64:
This should at least show us in the jenkins job output the full output
of `snapcraft push` and help identify failures to release or failures to
upload.

It also guards us against future changes in the output of
`snapcraft push`.

This will help in debugging the real failure behind jobs such as https://jenkins.edgexfoundry.org/view/Snap/job/edgex-go-snap-master-stage-snap/359/ even though it may not actually solve the build failure as we will be able to see the full output from `snapcraft push` without it going through grep.

Also I didn't deploy this to the sandbox because the sandbox doesn't have the necessary credentials in order to accurately test the stage job, but if desired I can try to hack around that.

### DIFF
--- a/shell/edgexfoundry-snapcraft.sh
+++ b/shell/edgexfoundry-snapcraft.sh
@@ -75,10 +75,9 @@ case "$JOB_TYPE" in
         popd > /dev/null
         pushd /build > /dev/null
         snapcraft login --with /build/edgex-snap-store-login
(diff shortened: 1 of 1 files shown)


tonyespy (approved):
LGTM

jamesrgregg (approved):
LGTM

PR #409, opened 2019-06-13
Adding codecov.io token in merge builds

lranjbar:
Forgot to put in the token that the shell script needs for codecov.io


eb-oss (approved):

eb-oss (changes requested):
Sorry Lisa, but could you please change the commit message to be imperative? Looks good otherwise!

eb-oss (approved):

PR #423, opened 2019-06-21
use edgex-ui-go-settings to verify pipeline

ernestojeda:
There seems to be an issue with the `edgex-ui-clojure-settings` file. I am updating the build to use the `edgex-ui-go-settings` to verify that there are no issues with the actual pipeline. Will revert after.


lranjbar (approved):
:+1:

Issue #417, opened 2019-06-19
Modify the Jenkins JJB template for volume view


jamesrgregg:
The current view named volume should be modified to be named: 
edgex-volume

jamesrgregg:
[US4752](https://rally1.rallydev.com/#/detail/userstory/315469527108?fdp=true)

jamesrgregg:
@vishwassp-hcl-intel completed this work.

PR #428, opened 2019-06-21
Rename volume view to edgex-volume

jamesrgregg:
Signed-off-by: James Gregg <james.r.gregg@intel.com>

### DIFF
--- a/jjb/common-views.yaml
+++ b/jjb/common-views.yaml
@@ -64,3 +64,12 @@
     filter-queue: false
     recurse: false
     regex: '.*edinburgh-stage.*'
+
+- view:
+    name: edgex-volume
+    description: 'List of edgex-volume jobs'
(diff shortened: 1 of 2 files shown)


jamesrgregg:
Fixes Issue #417

eb-oss:
I'm not sure this will work. A project-view is "^<project-name>-.\*", which means this will not display jobs named "docker-edgex-volume". Either adding the view in edgex-volume.yaml (where it would pick up the project name as defined there) or using a custom view so that it can be defined as "edgex-volume-.*" (without the '^' forcing it to only match at the beginning of the line) might be better solutions.

eb-oss:
jjb-deploy *edgex-volume*

jamesrgregg:
backing out these changes and push differently

PR #431, opened 2019-06-24
Add device-snmp-go staging jobs

lranjbar:

### DIFF
--- a/jjb/device/device-snmp-go.yaml
+++ b/jjb/device/device-snmp-go.yaml
@@ -18,3 +18,7 @@
       - '{project-name}-{stream}-merge-pipeline'
       - '{project-name}-pipeline-webhooks':
         status-context: '{project-name}-pipeline-webhooks'
+      - '{project-name}-{stream}-stage-go':
+          post_build_script: !include-raw-escape: shell/edgexfoundry-device-snmp-go-docker-push.sh
+      - '{project-name}-{stream}-stage-go-arm':
+          post_build_script: !include-raw-escape: shell/edgexfoundry-device-snmp-go-docker-push.sh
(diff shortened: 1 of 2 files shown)


eb-oss (changes requested):
A few notes here:
1) We don't push to release. Even as we work toward automated releases, they should move artifacts from staging to release.
2) Is there no way to make this more generic? I believe you should be able to pick up the project name from the environment; that would allow this to not be a single-use script, but instead be used by any other project that would like to go through the same actions.
3) Related to making it more generic, can the docker push and Nexus push be separated into two scripts?

lranjbar:
This pull request was created in response that we didn't have the proper release artifacts for Edinburgh.
What I've done is copy the same script pattern that exist for all the other jobs that we have release artifacts for. 

I recommend that we put the refactor for these scripts in the Fuji release. This is an issue we can bring up in the working group as well as part of the release retrospective.

Here is a list of all the scripts that are set up this way:
- https://github.com/edgexfoundry/ci-management/blob/master/shell/core-config-seed-go-docker-push.sh
- https://github.com/edgexfoundry/ci-management/blob/master/shell/edgexfoundry-device-modbus-go-docker-push.sh
- https://github.com/edgexfoundry/ci-management/blob/master/shell/edgexfoundry-device-mqtt-go-docker-push.sh
- https://github.com/edgexfoundry/ci-management/blob/master/shell/edgexfoundry-device-random-go-docker-push.sh
- https://github.com/edgexfoundry/ci-management/blob/master/shell/edgexfoundry-edgex-ui-go-push.sh
- https://github.com/edgexfoundry/ci-management/blob/master/shell/edgexfoundry-go-docker-push.sh
- https://github.com/edgexfoundry/ci-management/blob/master/shell/security-api-gateway-go-docker-push.sh
- https://github.com/edgexfoundry/ci-management/blob/master/shell/security-secret-store-go-docker-push.sh

eb-oss (approved):
Fair enough. I'll merge this as-is, and we can discuss a permanent solution for Fuji at the next Devops WG meeting.

PR #432, opened 2019-06-24
Add device-grove-c stage jobs

lranjbar:

### DIFF
--- a/jjb/device/device-grove-c.yaml
+++ b/jjb/device/device-grove-c.yaml
@@ -30,3 +30,7 @@
           build-node: ubuntu18.04-docker-arm64-4c-2g
           docker_build_args: '-f scripts/Dockerfile.alpine-3.9'
           docker_name: docker-device-grove-c-arm64
+      - '{project-name}-{stream}-stage-docker-arm':
+          build-node: ubuntu18.04-docker-arm64-4c-2g
+          docker_build_args: '-f scripts/Dockerfile.alpine-3.9'
+          docker_name: docker-device-grove-c-arm64
--- a/jjb/edgex-templates-docker.yaml
+++ b/jjb/edgex-templates-docker.yaml
@@ -189,6 +189,54 @@
       - shell: '{post_build_script}'
 
 - job-template:
+    name: '{project-name}-{stream}-stage-docker'
+    # Job template for Docker daily staging jobs
+    #
+    # The purpose of this job template is to stage a docker image for potential
+    # release.
+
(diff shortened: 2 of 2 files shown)


jamesrgregg (dismissed):
LGTM

jjb/edgex-templates-docker.yaml
```diff
@@ -188,6 +188,54 @@
       # Execute any post build clean-up or testing
       - shell: '{post_build_script}'
 
+- job-template:
+    name: '{project-name}-{stream}-stage-docker'
+    # Job template for Docker merge jobs
```
eb-oss: These comments are specific to merge jobs.

jjb/edgex-templates-docker.yaml
```diff
@@ -188,6 +188,54 @@
...
+      - shell: '{post_build_script}'
+      # Provided all steps have already passed, push the docker image
+      - shell: !include-raw: ../shell/docker-push.sh
+
+    triggers:
+      # no reason to add lf-infra-github-pr-trigger here since it doesn't
+      # currently work for merge / push
+      - github
+      - pollscm:
+          cron: '0 18 * * *'
```
eb-oss, line 237: This looks like it would run on merges and on the cron schedule. Is that what we want for staging?

jjb/device/device-grove-c.yaml
```diff
@@ -30,3 +30,7 @@
           build-node: ubuntu18.04-docker-arm64-4c-2g
           docker_build_args: '-f scripts/Dockerfile.alpine-3.9'
           docker_name: docker-device-grove-c-arm64
+      - '{project-name}-{stream}-stage-docker-arm':
```
eb-oss, line 33: Is there any intent to include non-arm builds? The list that Jim White provided me with for release included non-arm images for device-grove-c which are not actually being built.

eb-oss (changes requested):

jjb/edgex-templates-docker.yaml
lranjbar: Currently all the staging jobs run once a day on this same schedule.

jjb/device/device-grove-c.yaml
lranjbar: No, device-grove-c only supports the Raspberry Pi 3+ which is arm only. So we are only going to have arm images for this repo.

eb-oss (approved):

PR #434, opened 2019-06-26
Add docker build to device-snmp staging jobs

lranjbar:
It looks like the current staging jobs are failing because we aren't building docker images earlier in the job. I'm running these changes on the sandbox but it is hard to tell if they will work on production. (Failures happen on sandbox because of mocked out credentials)


jamesrgregg (approved):
LGTM

PR #444, opened 2019-07-10
Remove centos-7-arm64 builds from packer.yaml

eb-oss:
These are not being used, and the Vexxhost images are not working
properly, causing the jobs to always fail.


jamesrgregg (approved):
LGTM

PR #476, opened 2019-09-04
Remove invalid timeout setting

### LINKED ISSUE #450: Performance-test-master does not time out properly
@cherrycl Per this line, it appears that you intended the job to timeout after 60 minutes:
https://github.com/edgexfoundry/ci-management/blob/8e5e774255c5d754e047df5e0ce011f724ab597b/jjb/performance-test/performance-test.yaml#L17

However, because this is a pipeline job, this setting is not part of the config. I believe you'll need to add a timeout in the Jenkinsfile.

cherrycl:
Remove the timeout setting which can't use on a pipeline job.
Signed-off-by: Cherry Wang <cherry@iotechsys.com>

Fix #450

### DIFF
--- a/jjb/performance-test/performance-test.yaml
+++ b/jjb/performance-test/performance-test.yaml
@@ -13,8 +13,7 @@
           branch: 'master'
 
     jobs:
-      - 'performance-test-{stream}':
-          build-timeout: 60
+      - 'performance-test-{stream}'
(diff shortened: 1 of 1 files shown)


eb-oss (approved):

lranjbar (approved):
The timeout was moved to the Jenkinsflie

PR #493, opened 2019-09-27
Drop delhi snap jobs

anonymouse64:
Delhi is no longer supported as a release, so we can stop building these jobs.

Additionally, the jobs have been failing for some time due in part to a bug with the packaging that specified the master branch for docker-edgex-mongo, when it should have specified the delhi branch instead. Rather than fix that bug, we'll just stop builds for this.

### DIFF
--- a/jjb/edgex-go/edgex-go-snap.yaml
+++ b/jjb/edgex-go/edgex-go-snap.yaml
@@ -8,10 +8,6 @@
       - 'master':
           branch: 'master'
           snap-channel: latest/edge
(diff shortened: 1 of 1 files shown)


jamesrgregg (dismissed):
LGTM

eb-oss (dismissed):
Approved. Please rebase with master, and then we can merge.

anonymouse64:
@MightyNerdEric rebased

eb-oss (approved):

PR #511, opened 2019-10-28
Update JJBs for Fuji release

lranjbar:

### DIFF
--- a/jjb/consul/edgex-consul.yaml
+++ b/jjb/consul/edgex-consul.yaml
@@ -7,9 +7,11 @@
     stream:
       - 'master':
           branch: master
+      - 'fuji':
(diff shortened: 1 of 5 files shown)


jamesrgregg (dismissed):
LGTM

soda480 (dismissed):
LGTM

eb-oss (dismissed):
Looks fine. Now that all streams are using the same pre_build_script, build_script, and go-root in device-snmp-go and edgex-ui-go, those could be moved to the project level, so that they don't need to be repeated for every stream.

I'll approve as-is, but since this needs a rebase anyway, it may be worthwhile to fix that as well.

eb-oss (approved):

PR #519, opened 2019-11-11
jjb/{device,ui}: add fuji branch for snaps

anonymouse64:
These snaps should also have fuji jobs running for them.

I also filed https://jira.linuxfoundation.org/servicedesk/customer/portal/2/IT-18236 to get the tracks setup from the LF on the snap store.


anonymouse64:
CC @lranjbar @jamesrgregg

eb-oss (approved):

PR #571, opened 2020-02-11
edgex-go swagger fixes

ernestojeda:
created new script to wrap multiple call to publish different API versions.
This will make it easier to migrate this script to pipelines in the future.


ernestojeda:
recheck

ernestojeda:
Build failed due to python error. Trying recheck

soda480 (approved):
Looks good - I like the DryRun capability. Only question I have is why the OWNER and OASVERSION are hard coded as opposed to derived from the environment.

shell/toSwaggerHub.sh
```diff
@@ -1,17 +1,41 @@
...
+                    -H "Content-Type:application/yaml" \
+                    -d "${apiContent}"
+            else
+                echo "[toSwaggerHub] Dry Run enabled...Simulating upload"
+                echo "curl -X POST https://api.swaggerhub.com/apis/${owner}/${apiName}?isPrivate=${isPrivate}&version=${apiVersion}&oas=${oasVersion}&force=true"
+            fi
+
+        done
+    else
+        echo "Could not API Folder [${apiPath}]. Please make sure the API version exists..."
```
jamesrgregg: are we missing the word "find" in this echo statement?

jamesrgregg (approved):
Other than the missing word in the echo statement.. LGTM

shell/toSwaggerHub.sh
ernestojeda: sure. I will add it.

ernestojeda:
recheck

PR #582, opened 2020-02-21
Remove unused nodes, update ubuntu-18.04 base

eb-oss:
This removes all ubuntu-16.04 and snapcraft nodes, none of which are
in use currently. It also removes unused versions of current nodes,
such as the centos7-docker-2c-1g node and the ubuntu-18.04-builder
packer build.

Finally, this also modifies the ubuntu-18.04-arm64 base image used in
the packer build. The 2019-12-11 image set in common-packer has seen
issues in other projects, so we will go back to a known-good image.

Issue: LF-Jira RELENG-2751


eb-oss:
Quick bump here. This still needs review. Please let me know if there are any questions.

jamesrgregg (approved):
LGTM

PR #585, opened 2020-03-02
Publish v1 API to Swagger and remove some variables

### LINKED ISSUE #579: The V1 APIs on edgex-go also need to publish to swaggerhub
I found the edgex-go [jenkins build](https://jenkins.edgexfoundry.org/view/EdgeX%20Foundry%20Project/job/edgexfoundry/job/edgex-go/job/master/4/consoleFull) only execute publishing v2 APIs to SwaggerHub.
The v1 APIs also need to publish to SwaggerHub by jenkins job.

- V1 APIs put on [edgex-go/api/openapi/v1](https://github.com/edgexfoundry/edgex-go/tree/master/api/openapi/v1)
- The [shell script](https://github.com/edgexfoundry/ci-management/blob/master/shell/edgex-publish-swagger.sh) didn't see any v1 API input.

cherrycl:
1. Remove isPrivate and API_VERSION variables. The API_VERSION will get from API file directly, and set the API to public when sending the API to SwaggerHub.
2. Publish v1 API to SwaggerHub.
Signed-off-by: Cherry Wang <cherry@iotechsys.com>
Fix #579

### DIFF
--- a/shell/edgex-publish-swagger.sh
+++ b/shell/edgex-publish-swagger.sh
@@ -9,8 +9,8 @@ if [ -z "$ARCH" ] || [ "$ARCH" != "arm64" ] ; then
     # NOTE: APIKEY needs to be a pointer to a file with the key. This will need to be set locally from your environment or from Jenkins
     APIKEY_VALUE=`cat $APIKEY`
 
-    API_VERSION=${API_VERSION:-2.0.0}
-    API_FOLDER=${API_FOLDER:-v2}
+    # Upload both API versions,v1 and v2, to SwaggerHub
+    API_FOLDERS="v1 v2"
     SWAGGER_DRY_RUN=${SWAGGER_DRY_RUN:-false}
 
     SCRIPTS_ROOT="$( cd "$(dirname "$0")" >/dev/null 2>&1 ; pwd -P )"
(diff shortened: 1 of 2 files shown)


cherrycl:
@ernestojeda Would you please help to review this PR? Thank you.

shell/edgex-publish-swagger.sh
```diff
@@ -9,19 +9,24 @@ if [ -z "$ARCH" ] || [ "$ARCH" != "arm64" ] ; then
     # NOTE: APIKEY needs to be a pointer to a file with the key. This will need to be set locally from your environment or from Jenkins
     APIKEY_VALUE=`cat $APIKEY`
 
-    API_VERSION=${API_VERSION:-2.0.0}
-    API_FOLDER=${API_FOLDER:-v2}
+    # API_VERSION=${API_VERSION:-2.0.0}
```
eb-oss: Is there a reason to leave these lines (12, 13, 22) commented out? If they're not necessary, they should be removed. If you're not sure if they should be removed, this should be tested more thoroughly. Keep in mind that removed lines will be kept in git history, in the case that they need to be restored later.

eb-oss (changes requested):

shell/edgex-publish-swagger.sh
cherrycl: @MightyNerdEric Thanks for your advice. I removed the lines commented out.

shell/toSwaggerHub.sh
```diff
@@ -5,11 +5,9 @@ echo "--> toSwaggerHub.sh"
 publishToSwagger() {
     apiKey=$1
     apiFolder=$2
-    apiVersion=$3
-    oasVersion=$4
-    isPrivate=$5
```
ernestojeda, line 10: @cherrycl after talking with @lenny-intel I don't think removing `isPrivate` is the correct behavior.

ernestojeda (changes requested):
Everything looks OK except the isPrivate. I think this needs to be discussed in the core WG before this change goes live.

ernestojeda:
I am adding @lenny-intel to this review because he had some objections about removing `isPrivate`

cloudxxx8:
@ernestojeda and @lenny-intel let's discuss in tomorrow QA WG meeting about the `isPrivate` topic.
If you can't join the call, we can still discuss in this issue.

cherrycl:
Based on the discussion on this week's QA meeting, reset the API to private when publishing API to SwaggerHub.
@ernestojeda @lenny-intel Please help to review this PR again. Thanks.

lenny-goodell (commented):
My concerns have been addressed. I'll leave the final approval to those that better understand these scripts.

ernestojeda (approved):
👍 Looks good after the changes

cloudxxx8:
@MightyNerdEric Cherry has addressed your change request.  Could you please review again?  Thanks.

eb-oss (approved):
This just needs a rebase now.

cherrycl:
Rebased.

PR #617, opened 2020-07-13
snap: update for new snapcraft, fix old bugs

### LINKED ISSUE #369: Snap builds should use unbuffer for better output logging
See https://forum.snapcraft.io/t/fixing-out-of-order-message-output-when-piping-snapcrafts-output-to-file/9176 for full explanation, but basically we should update the docker build script for the snap to use unbuffer so that the log looks a bit better as oftentimes on jenkins logs the error gets output to stderr much earlier than the corresponding output at the same time on stdout, hence one has to do some backtracking when debugging the logs.

Originally filed as https://github.com/edgexfoundry/edgex-go/issues/1018 but now since the snap build scripts are here, migrating to this repo

### LINKED ISSUE #517: snap build job doesn't properly fail on incorrect shas of downloaded snaps
See for example the log from [this job](https://jenkins.edgexfoundry.org/job/edgex-go-snap-fuji-verify-snap/9/console):

```
22:26:13 [91msha512sum: WARNING: 1 computed checksum did NOT match
22:26:13 [0mcore.snap: FAILED
22:26:13 Parallel unsquashfs: Using 4 processors
22:26:13 [91mRead on filesystem failed because EOF
22:26:13 read_uids_guids: failed to read id index table
22:26:13 FATAL ERROR:failed to uid/gid table
22:26:13 [0mcore18.snap: OK
```
the job still goes on to try to run and fails for other unknown reasons.

anonymouse64:
This PR fixes the broken CI jobs for edgex snaps since upstream snapcraft now requires to use `snap pack` instead of mksquashfs. There was not an issue filed for this, perhaps @tonyespy can file one if needed.

It also fixes the long standing issue https://github.com/edgexfoundry/ci-management/issues/369 by installing the expect package in the docker container used to build the snap and using the associated `unbuffer` command to prevent output from getting mangled by docker stdout streaming.

Finally, it also fixes other issues we have had where if the download for the core, core18, or snapcraft snaps fail and the hash doesn't match, we now fail the build much earlier using `&&` instead of `;` between commands. An issue for this type of issue was filed at https://github.com/edgexfoundry/ci-management/issues/517.

This can be tested by checking out master branch of edgex-go and running the edgexfoundry-snapcraft.sh script from this repo from inside the edgex-go repo, i.e.

```bash
$ git clone https://github.com/edgexfoundry/edgex-go.git
$ cd edgex-go
$ # git clone this PR
$ /home/user/git/ci-management/shell/edgexfoundry-snapcraft.sh 
... should produce working output and no snapcraft error trace like we see in https://github.com/edgexfoundry/edgex-go/pull/2613#issuecomment-657112589
```

Not sure what labels to use for this PR, the PR labels are not documented in https://github.com/edgexfoundry/ci-management/blob/master/.github/CONTRIBUTING.md

## PR Checklist

- [ ] The commit message follows our guidelines: https://github.com/edgexfoundry/ci-management/blob/master/.github/CONTRIBUTING.md
- [ ] Added labels
## PR Type

- [x] Bugfix
- [x] Feature
- [x] Refactoring (no functional changes, no api changes)

## Issue Number:
* Fixes #369
* Fixes #517

## Sandbox Testing
Test Links :

Didn't test in the sandbox since it can be tested locally as mentioned above. I would have to dig up my old instructions on how to do that before I can test in the sandbox

### DIFF
--- a/shell/edgexfoundry-snapcraft.sh
+++ b/shell/edgexfoundry-snapcraft.sh
@@ -70,29 +70,29 @@ case "$JOB_TYPE" in
     "stage")
         # Stage jobs build the snap locally and release it
         pushd /build > /dev/null
-        snapcraft clean
-        snapcraft
+        unbuffer snapcraft clean
+        unbuffer snapcraft
         popd > /dev/null
         pushd /build > /dev/null
-        snapcraft login --with /build/edgex-snap-store-login
+        unbuffer snapcraft login --with /build/edgex-snap-store-login
         # Push the snap up to the store and release it on the specified
         # channel
-        snapcraft push "$SNAP_NAME"*.snap --release "$SNAP_CHANNEL" 
+        unbuffer snapcraft push "$SNAP_NAME"*.snap --release "$SNAP_CHANNEL" 
         # Also force an update of the meta-data
-        snapcraft push-metadata "$SNAP_NAME"*.snap --force
+        unbuffer snapcraft push-metadata "$SNAP_NAME"*.snap --force
         popd > /dev/null
     ;;
     "release")
         # Release jobs will promote an already built snap revision
         # in the store to a channel.
-        snapcraft login --with /build/edgex-snap-store-login
-        snapcraft release "$SNAP_NAME" "$SNAP_REVISION" "$SNAP_CHANNEL"
+        unbuffer snapcraft login --with /build/edgex-snap-store-login
+        unbuffer snapcraft release "$SNAP_NAME" "$SNAP_REVISION" "$SNAP_CHANNEL"
     ;;
     *)
         # Do normal build and nothing else to verify the snap builds
         pushd /build > /dev/null
-        snapcraft clean
-        snapcraft
+        unbuffer snapcraft clean
+        unbuffer snapcraft
         popd > /dev/null
     ;;
 esac
@@ -150,7 +150,7 @@ RUN apt-get update && \
     apt remove --yes --purge curl jq squashfs-tools && \
     apt-get autoclean --yes && \
     apt-get clean --yes && \
-    apt-get install snapd --yes
+    apt-get install snapd expect --yes
 
 
 # The upstream dockerfile just uses this file locally from the repo since it's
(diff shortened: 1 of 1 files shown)


eb-oss (changes requested):
Can you squash the 3 commits down to one?

anonymouse64:
@MightyNerdEric the individual commits fix different issues, would you rather I submit 3 separate PR's ?

ernestojeda:
@anonymouse64 These scripts are no longer used to push snaps. All snap scripts have been moved to ci-build-images. You can find the corresponding script/docker image here:

https://github.com/edgexfoundry/ci-build-images/blob/snapcraft-builder/Dockerfile
https://github.com/edgexfoundry/ci-build-images/blob/snapcraft-builder/entrypoint.sh

Are you able to make those changes in in the ci-build-images repo?

ernestojeda (changes requested):
See my comment. I will work to remove these files to avoid the confusion.

tonyespy (approved):
Tested build, which now works, and reviewed the code changes.

anonymouse64:
@ernestojeda how does one test the changes to that repo locally without using the sandbox? I'm happy to submit the changes there instead but I don't know how the pipelines jobs work so I don't know how to reproduce what jenkins does locally with that repo

ernestojeda:
> @ernestojeda how does one test the changes to that repo locally without using the sandbox? I'm happy to submit the changes there instead but I don't know how the pipelines jobs work so I don't know how to reproduce what jenkins does locally with that repo

@anonymouse64  Yeah, it will be pretty much impossible to test this locally, but It is definitely possible to test this on the sandbox. However, it will be pretty difficult to pull off. If it's OK with you, would you mind if the DevOps team took these changes over and integrated them into the ci-build-images repository? We could then demo this in an upcoming WG meeting to show the testing process.

anonymouse64:
@ernestojeda sounds fine to me, only thing would be that all the snap PR jobs are currently broken, so if you could at least commit / merge https://github.com/edgexfoundry/ci-management/pull/617/commits/7c46d5af2ece321a4a06a2844fbdb0228b5dc6e7 ASAP that would be appreciated. The other commits are not needed ASAP so are fine to re-use for demo purposes, etc. 

With that in mind do you still want me to submit PR's against ci-build-images or will you take care of that?

ernestojeda:
I will take care of that.

PR #625, opened 2020-08-24
feat: Add new centos node of flavor v1-standard-8

ernestojeda:
Add new node type to help speed up Snap builds. I did some testing last week with this node type and edgex-go Snap builds go from ~30 minutes down to ~16 minutes.

## PR Checklist

- [x] The commit message follows our guidelines: https://github.com/edgexfoundry/ci-management/blob/master/.github/CONTRIBUTING.md
- [x] Added labels
## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Refactoring (no functional changes, no api changes)

## Sandbox Testing
Test Links :

### DIFF
--- a/jenkins-config/clouds/openstack/Primary/centos7-docker-8c-8g.cfg
+++ b/jenkins-config/clouds/openstack/Primary/centos7-docker-8c-8g.cfg
@@ -1,0 +1,2 @@
+IMAGE_NAME=ZZCI - CentOS 7 - docker - x86_64 - 20200505-094219.351
+HARDWARE_ID=v1-standard-8


jamesrgregg (approved):
👍 - impressive optimization @ernestojeda

eb-oss (commented):
As I mentioned before, we need to move off of v1 nodes. We can do that as a separate change, so this one is okay to add. But I'm wondering: do you want to test out those changes before we make them, so you can check on time losses/gains for different nodes? Since the v1s are not exactly equivalent to the v2s (as I posted in the devops channel, v2s tend to be lighter on cpu, heavier on ram) we might not see the same gains just by switching to, say, v2-standard-8.

ernestojeda:
I will need some help to get the v2 configuration working. I tried a few different v2 flavors and could not get any to work. I know you mentioned v2 may require some volume configuration do you have examples of that?

eb-oss:
> I will need some help to get the v2 configuration working. I tried a few different v2 flavors and could not get any to work. I know you mentioned v2 may require some volume configuration do you have examples of that?

Yes, you can see the change I made for the Acumos project here: https://github.com/acumos/ci-management/commit/1906e454c525742b4a57a11e7489f92bd45b2206
The one thing I'd say is that 40gb is a safer volume size. I will probably start defaulting to that rather than 20 in the future.

ernestojeda:
I will work to get v2 working out of the box then. Seems pretty straight forward.

eb-oss (approved):

PR #700, opened 2025-05-26
ci: Add sandbox maven setting files

cherrycl:
## PR Checklist

- [ ] The commit message follows our guidelines: https://github.com/edgexfoundry/ci-management/blob/master/.github/CONTRIBUTING.md
- [ ] Added labels
## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Refactoring (no functional changes, no api changes)

## Sandbox Testing
Test Links :

### DIFF
--- a/jenkins-config/managed-config-files/mavenSettings/sandbox-settings/serverCredentialMappings.sandbox.yaml
+++ b/jenkins-config/managed-config-files/mavenSettings/sandbox-settings/serverCredentialMappings.sandbox.yaml
@@ -1,0 +1,12 @@
+---
+serverCredentialMappings:
+  - serverId: "nexus3.edgexfoundry.org:10001"
+    credentialsId: "docker"
+  - serverId: "nexus3.edgexfoundry.org:10002"
+    credentialsId: &cred "sandbox"
+  - serverId: "nexus3.edgexfoundry.org:10003"
+    credentialsId: *cred
+  - serverId: "nexus3.edgexfoundry.org:10004"
+    credentialsId: *cred
+  - serverId: "docker.io"
+    credentialsId: edgex_dockerhub_creds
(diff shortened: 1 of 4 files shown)


eb-oss (changes requested):
A couple of notes here:

1. I don't think the codecov token is necessary. Unless your testing has shown that it's required, this should be removed.
2. Since these tokens are specific to sandbox testing, you should actually flip the [...]Mappings.yaml and Mappings.sandbox.yaml contents. Put the actual contents in the .sandbox file, and the other one can be symlinked to the dummy sandbox file, since we don't want that to have any content on the prod system.

cherrycl:
@eb-oss Thanks for your comment. The PR has been updated.

jenkins-config/managed-config-files/mavenSettings/sandbox-settings/serverCredentialMappings.sandbox.yaml
```diff
@@ -0,0 +1,12 @@
+---
+serverCredentialMappings:
+  - serverId: "nexus3.edgexfoundry.org:10001"
+    credentialsId: "docker"
+  - serverId: "nexus3.edgexfoundry.org:10002"
+    credentialsId: &cred "sandbox-settings"
```
eb-oss: This should just be `sandbox` rather than `sandbox-settings`.

eb-oss (changes requested):

jenkins-config/managed-config-files/mavenSettings/sandbox-settings/serverCredentialMappings.sandbox.yaml
cherrycl: Updated.

eb-oss (approved):

